PR #6, opened 2021-06-29
Check if line exists before setting cursor

ten3roberts:
There is currently an issue generating error `Cursor position outside buffer` in `aerial/window.lua:278` when using a global aerial window and switching to other windows.

When the cursor moves to another buffer, the last position that was selected for the aerial window is set. When the cursor previously was in a window with fewer symbols, this causes the line to not exist because symbols may publish after `CursorMoved` is triggered.

This fix checks if the line of the last selected symbol and line exists before jumping to it. The correct line will be selected when the symbols for the current buffer are rendered and the cursor moves.

### DIFF
--- a/lua/aerial/window.lua
+++ b/lua/aerial/window.lua
@@ -4,10 +4,13 @@ local loading = require 'aerial.loading'
 local render = require 'aerial.render'
 local util = require 'aerial.util'
 
+local api = vim.api
+local fn = vim.fn
+
 local M = {}
 
 local function create_aerial_buffer(bufnr)
-  local aer_bufnr = vim.api.nvim_create_buf(false, true)
+  local aer_bufnr = api.nvim_create_buf(false, true)
 
   util.go_buf_no_au(aer_bufnr)
   if config.default_bindings then
@@ -16,7 +19,7 @@ local function create_aerial_buffer(bufnr)
         keys = {keys}
       end
       for _,key in ipairs(keys) do
-        vim.api.nvim_buf_set_keymap(aer_bufnr, 'n', key, cmd, {silent = true, noremap = true})
+        api.nvim_buf_set_keymap(aer_bufnr, 'n', key, cmd, {silent = true, noremap = true})
       end
     end
     map('<CR>', "<cmd>lua require'aerial'.select()<CR>")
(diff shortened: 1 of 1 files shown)


stevearc:
The change looks good, but could I get a bit more information about the repro? I tried it myself and managed to get a similar error due to some conflicts with nvim-treesitter-context (fixed in 958855d404160897b68291f76038075a1eea1128), but I don't think that it was the same issue you're reporting because it also resulted in clearing the aerial buffer entirely. I'll likely merge this either way, but I want to see it myself because I don't think we should be updating the cursor position before the symbols have been swapped. I want to make sure that the root cause isn't something nastier going on that needs more fixing.

Is there anything more to it than switching from a buffer with fewer symbols to a buffer with more symbols (with the cursor at the bottom)? Does it happen consistently? What language server?

ten3roberts:
I am sorry, maybe I should have been more clear on this.

Steps to reproduce:
- Open buffer and attach `rust-analyzer`
- Open Aerial, with `close_behavior = 'global'`
- Do `:Telescope find_files`
- The Aerial window is cleared, but doesn't redraw before error is shown

Screenshot from 2021-06-30 07-52-17

It sometimes happens with the quickfix, but it always occurs when using Telescope. I think it may be because Telescope moves the cursor to the prompt.

The same error also occurs a second time when closing Telescope.

When using FZF, the issue does not occur when initially opening the window, but occurs when closing it and returning to the previous window with code, but now with an empty Aerial window.

Screenshot from 2021-06-30 07-58-37

Another fix for this issue would be to not clear the Aerial buffer, although this fix would change the default behavior of showing an empty window when a `normal` window without lsp is focused, and would falsely show symbols for a code window then there aren't any. Furthermore, this could still cause the same issue of positioning the cursor outside the buffer to occur in other circumstances.

My belief is just letting the cursor be when it is not valid to move it to the new position is the least intrusive fix and isn't noticed during the time the issue occurs, I.e when using telescope, because when telescope closes and symbols are published it is positioned correctly.

stevearc:
Thanks for the detailed writeup! So it turns out this *is* similar to the issue I found, but for some reason I still can't repro it. The change I made yesterday (958855d) should hopefully fix the issue with Telescope, but it won't do anything for the quickfix. Since I can't reproduce the issue to look at it closer, I'm inclined to agree with you that the best thing to do is to let it pass silently. LMK if you see any weirdness in the future.

PR #19, opened 2021-11-23
feat: added option to automatically close aerial on select (close_on_select)

axieax:
This PR adds the ability to configure aerial to automatically close the window upon selecting/jumping to a symbol. Please note that vimdocs will need to be generated for this :))

### DIFF
--- a/lua/aerial/config.lua
+++ b/lua/aerial/config.lua
@@ -79,6 +79,9 @@ local default_options = {
   -- Run this command after jumping to a symbol (false will disable)
   post_jump_cmd = "normal! zz",
 
+  -- If close_on_select is true, aerial will automatically close after jumping to a symbol
+  close_on_select = false,
+
   lsp = {
     -- Fetch document symbols when LSP diagnostics change.
     -- If you set this to false, you will need to manually fetch symbols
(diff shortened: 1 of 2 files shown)


stevearc:
Thanks for the PR! I like the functionality, but this feels more like a keymap than a config option to me. I could see someone wanting to do either "jump to loc" or "jump to loc and close aerial" depending on what they're currently doing. It's pretty easy to override the default behavior by adding this to `ftplugin/aerial.vim`:
```vim
nnoremap <buffer> <CR> <cmd>lua require'aerial'.select()<CR><cmd>AerialClose<CR>
```

What do you think of adding a default keybinding that does this, to make it more discoverable? Or possibly adding this as documentation so it's easy to find?

axieax:
Hi thanks for the quick response! I feel like having it as a config option is easier to work with (personally I prefer to keep my config settings in one place rather than in separate files, which makes it easier to maintain/tweak stuff later). This also means that users don't have to manually remap every jump to symbol mapping manually as well (CR, braces, square brackets, splits etc.), which can be rather annoying when more default keybindings are released/changed as well.

stevearc:
Fair enough. Thanks again!

PR #24, opened 2021-11-29
feat: Add basic dart treesitter support

GenesisTMS:
It is very basic list of dart TS nodes. Will add more whenever i find
missing one.


stevearc:

PR #26, opened 2021-12-01
Add lualine component to render current symbol

ikalnytskyi:
Lualine is a widely used NeoVim statusline plugin. This patch adds a
lualine component that renders current symbol. It supports two rendering
modes:

1. When `hierarchy` is set to `true`, it renders the whole path to the
   current symbol including parent nodes.

2. Otherwise, it renders the current symbol only.

The component also respects `icons_enabled` option that can turn on/off
symbol icons rendering.

### DIFF
--- a/lua/lualine/components/aerial.lua
+++ b/lua/lualine/components/aerial.lua
@@ -1,0 +1,116 @@
+-- ## Usage
+--
+-- Just specify "aerial" component in one of your lualine sections.
+-- For example:
+--
+--   require("lualine").setup({
+--     sections = {
+--       lualine_x = { "aerial" },
+--     },
+--   })
+--
+-- ## Options
+--
+-- *sep* (default: " ⟩ ")
+--   The separator to be used to separate symbols in status line.
+--
+-- *depth* (default: nil)
+--   The number of symbols to render top-down. In order to render only 'N' last
+--   symbols, negative numbers may be supplied. For instance, 'depth = -1' can
+--   be used in order to render only current symbol.
+--
+-- *dense* (default: false)
+--   When 'dense' mode is on, icons are not rendered near their symbols. Only
+--   a single icon that represents the kind of current symbol is rendered at
+--   the beginning of status line.
+--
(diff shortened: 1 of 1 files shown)


ikalnytskyi:
I'm also thinking about third rendering mode, something like:

```
{last_symbol_icon} {parent}.{parent}.{child}
```

:thinking:

lua/lualine/components/aerial.lua
```diff
@@ -0,0 +1,36 @@
+local M = require("lualine.component"):extend()
+local aerial = require("aerial")
+
+local default_options = {
+  sep = " ⟩ ",
```
ikalnytskyi, line 5: **n.b.** I can't call it `separator` since `separator` is reserved by `lualine`.

lua/lualine/components/aerial.lua
```diff
@@ -0,0 +1,36 @@
...
+  hierarchy = true,
```
ikalnytskyi, line 6: **Q:** What would be a better name?

stevearc: I think hierarchy is good

stevearc:
LGTM! Could you add a comment at the top of the file showing how one would add it to the lualine config?

ikalnytskyi:
> Could you add a comment at the top of the file showing how one would add it to the lualine config?

Will do. I also want to try some other configuration first. I think I'll have some time to try it out tomorrow.

ikalnytskyi:
Hey @stevearc,

I just updated the patch. Can you please take a look once you have a moment? It may become a bit harder to read, yet I think it will cover most use cases out there.

ikalnytskyi:
I'm not happy about option names. If you have better ideas, please let me know.

stevearc (changes requested):
Looks good! I think the option names are fine. Could you address the lint before merging?

ikalnytskyi:
I think I fixed them.

stevearc:

PR #30, opened 2021-12-10
Fix deprecated vim.lsp.diagnostic.get_count

ten3roberts:
As of version 0.7+ the vim.lsp.diagnostic API is deprecated.

This pull request fixes the deprecated api usage


ten3roberts:
This fixes issue #31

stevearc:
Unfortunately there's a bit more to do in order to correctly migrate away from `vim.lsp.diagnostic`. I took a stab at a complete fix and threw in a few more changes I've been wanting to make (migrating off of autocmds and limiting the scope to a single LSP client).

Thanks for the PR, though!

PR #46, opened 2022-01-14
Implement padchar configuration

reo101:
The the [`rpad`](https://github.com/stevearc/aerial.nvim/blob/1c78147a2aea5b00d8291b94158591dfa435ebaf/lua/aerial/util.lua#L5-L11) and [`lpad`](https://github.com/stevearc/aerial.nvim/blob/1c78147a2aea5b00d8291b94158591dfa435ebaf/lua/aerial/util.lua#L13-L18) functions have a third parameter for an (optional) padchar but in it isn't passed anywhere (omitted [here](https://github.com/stevearc/aerial.nvim/blob/de0170fce2242e3482bee8cb31858477cc456a4d/lua/aerial/bindings.lua#L84) and [here](https://github.com/stevearc/aerial.nvim/blob/1c78147a2aea5b00d8291b94158591dfa435ebaf/lua/aerial/render.lua#L86)). That leads to `" "` (whitespace) always being the padchar which (combined with a `listchar` for trailing spaces) also leads to situations like this:
image

In this PR, I've added a `padchar` option to the config for aerial that eliminates this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,11 @@ require("aerial").setup({
   -- This can be a function (see :help aerial-open-automatic)
   open_automatic = false,
 
+  -- The character that fills the empty space between the end of the
+  -- symbol name and the edge of the aerial window
+  -- (leaving this as "" will cause partial highlighting on current symbol)
+  padchar = " ",
+
   -- Set to true to only open aerial at the far right/left of the editor
   -- Default behavior opens aerial relative to current window
   placement_editor_edge = false,
--- a/lua/aerial/bindings.lua
+++ b/lua/aerial/bindings.lua
@@ -1,4 +1,5 @@
 local util = require("aerial.util")
+local config = require("aerial.config")
 local M = {}
 
 M.keys = {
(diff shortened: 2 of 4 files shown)


stevearc:
The purpose of padding out the whitespace at the end of the line is so that the line highlight is full-width. If your listchars setting causes the display to look bad, wouldn't a better solution be to change listchars for the aerial window? I think it would be reasonable to set a default that hides trailing spaces.

Would you actually want the aerial window to have something *other* than whitespace padding out the line?

reo101:
I agree, one can change the window-local `listchars` (maybe with an autocommand), and also yes - I think a padding of dashes or underscores would look nice too.

stevearc:
How about this as a solution then: I added a commit that disables most listchars by default in the aerial window. If you or anyone wants the padding characters to be different, you can modify those listchars with an autocmd, e.g.
```vim
autocmd FileType aerial setlocal listchars+=trail:-
```
Does that fix all of your issues?

reo101:
Yes, that would do it! Thanks a lot 😄

reo101:
Small update: `:remove()`-ing `listchars` from `opt_local` only cleara previously set local `listchars` but not the original ones.

stevearc:
Apologies! When I tested I was using a global listchars value that included options that *weren't* removed by aerial, so the remaining window-local value was sufficient to override the global defaults. Upon further testing, that doesn't work if, for example, all you have set is `trail:-`. I changed how the override is happening to be simpler and hopefully get rid of those edge cases.

reo101:
Thank you yet again! I can now confirm that it is working as expected. Have a great day and keep up the good work 🎉

PR #59, opened 2022-02-20
Add 'ignore' configuration options

b0o:
Hi! Thank you for this wonderful plugin, it's been a great addition to my workflow.

This PR adds options to configure aerial to ignore certain windows/buffers. Ignored windows/buffers change the behavior of aerial as follows:

- If `close_behavior` is `"global"`, focusing an ignored window/buffer will not cause the aerial window to update.

- If `open_automatic` is `true`, focusing an ignored window/buffer will not cause an aerial window to open.

- If `open_automatic` is a function, ignore rules have no effect on aerial window opening behavior; it's handled by the `open_automatic` function.

The primary reason I made these changes is that when `close_behavior` is `"global"`, focusing a special window like [nvim-tree](https://github.com/kyazdani42/nvim-tree.lua) caused the aerial window to become blank. This was rather jarring; I'd prefer for the aerial buffer to stay unchanged if I focus such windows. The new `ignore.buftypes = "special"` option resolves this issue.

I added additional configuration options to give the user maximal flexibility, as there may be cases where they want to ignore a certain filetype or the buffer of a certain plugin. In most cases, though, the default options should suit the user well.

#### Notes

- Somewhat similar functionality already existed with setting `open_automatic` to a function, however, this did not help with the issue I described above with `close_behavior = "global"`.
- For the `ignore.buftypes` option, the value `"special"` is a shortcut for `ignore.buftypes = { "acwrite", "help", "nofile", "nowrite", "quickfix", "terminal", "prompt" }`. Likewise, `ignore.wintypes = "special"` is a shortcut for `ignore.wintypes = { "autocmd", "command", "loclist", "popup" "preview" "quickfix", "unknown" }`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,44 @@ require("aerial").setup({
   -- If you have lspkind-nvim installed, aerial will use it for icons.
   icons = {},
 
+  -- Control which windows and buffers aerial should ignore.
+  -- If close_behavior is "global", focusing an ignored window/buffer will
+  -- not cause the aerial window to update.
+  -- If open_automatic is true, focusing an ignored window/buffer will not
+  -- cause an aerial window to open.
+  -- If open_automatic is a function, ignore rules have no effect on aerial
+  -- window opening behavior; it's entirely handled by the open_automatic
+  -- function.
+  ignore = {
+    -- Ignore unlisted buffers. See :help buflisted
+    unlisted_buffers = true,
+
+    -- List of filetypes to ignore.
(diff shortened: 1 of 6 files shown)


stevearc:
Sick! This is excellently and thoroughly done, thank you. 

Thanks for making SchemaStore.nvim, by the way! Always cool to run into the people behind the tools I use every day 😄

b0o:
Awesome, thanks for the quick merge!

PR #69, opened 2022-03-20
Add: Link to default bindings in README.md

eddiebergman:
Hi @stevearc,

Nice plugin :) I was originally trying to overwrite the key-bindings with the commands listed in the README, specifically replace `<CR>` to open the currently selected node in the AerialTree.

However there is no command I can see for this and I needed to see what the default bindings were doing. This just adds an extra link as the information of what to overwrite default bindings with seems to be missing.

Best,
Eddie

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ filtering.
 ## Default Keybindings
 
 The default keybindings in the aerial window. You can add your own in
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM. Thanks for the PR!

PR #70, opened 2022-03-22
feat: highlight group for non-current split line

b0o:
Adds a highlight group `AerialLineNC` which is applied to the portion of a split line highlight for the non-current window. Requires `highlight_mode="split_width"` to have any effect.

By default, `AerialLineNC` is linked to `AerialLine`, so users will not notice a difference from prior behavior unless they define their own `AerialLineNC` highlight group.

I chose the `NC` name because it seems to be the convention used by vim for highlights relating to non-current windows, for example [`:help hl-StatusLineNC`](https://neovim.io/doc/user/syntax.html#hl-StatusLineNC).

https://user-images.githubusercontent.com/21299126/159594894-978f6e85-fafd-4016-b5ae-0864d00e6f76.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,6 +540,10 @@ hi link AerialGuide Comment
 " You can set a different guide color for each level
 hi AerialGuide1 guifg=Red
 hi AerialGuide2 guifg=Blue
+
+" If highlight_mode="split_width", you can set a separate color for the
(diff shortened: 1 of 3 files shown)


stevearc:
Love it!

Issue #76, opened 2022-04-03
Error with `open_automatic` and folke/zen-mode.nvim


metalelf0:
**Describe the bug**

When using `folke/zen-mode.nvim`, and navigating between buffers, I get an error if aerial.nvim auto-open is enabled: `Error executing vim.schedule lua callback: ...m/site/pack/packer/start/aerial.nvim/lua/aerial/util.lua:52: Expected Lua number`

**System information**
- OS: macOS 12.3
- Neovim version: v0.7.0-dev+1345-gd89a80faf
- AerialInfo:

```
Aerial Info                                                                                                                                                                                                                                                        
-----------                                                                                                                                                                                                                                                        
Filetype: lua                                                                                                                                                                                                                                                      
Configured backends:                                                                                                                                                                                                                                               
  treesitter (supported) (attached)                                                                                                                                                                                                                                
  lsp (supported)                                                                                                                                                                                                                                                  
  markdown (not supported) [Filetype is not markdown]                                                                                                                                                                                                              
Show symbols: Class, Constructor, Enum, Function, Interface, Module, Method, Struct  
```

- Aerial config:
```lua
-- Call the setup function to change the default behavior
require("aerial").setup({
  -- Priority list of preferred backends for aerial.
  -- This can be a filetype map (see :help aerial-filetype-map)
  backends = { "treesitter", "lsp", "markdown" },

  -- Enum: persist, close, auto, global
  --   persist - aerial window will stay open until closed
  --   close   - aerial window will close when original file is no longer visible
  --   auto    - aerial window will stay open as long as there is a visible
  --             buffer to attach to
  --   global  - same as 'persist', and will always show symbols for the current buffer
  close_behavior = "auto",

  -- Set to false to remove the default keybindings for the aerial buffer
  default_bindings = true,

  -- Enum: prefer_right, prefer_left, right, left, float
  -- Determines the default direction to open the aerial window. The 'prefer'
  -- options will open the window in the other direction *if* there is a
  -- different buffer in the way of the preferred direction
  default_direction = "prefer_right",

  -- Disable aerial on files with this many lines
  disable_max_lines = 10000,

  -- A list of all symbols to display. Set to false to display all symbols.
  -- This can be a filetype map (see :help aerial-filetype-map)
  -- To see all available values, see :help SymbolKind
  filter_kind = {
    "Class",
    "Constructor",
    "Enum",
    "Function",
    "Interface",
    "Module",
    "Method",
    "Struct",
  },

[...]

PR #77, opened 2022-04-03
🐛 Don't crash on missing winid

metalelf0:
Fixes https://github.com/stevearc/aerial.nvim/issues/76.


stevearc:
LGTM. Thanks for the PR!

PR #83, opened 2022-04-18
feat: add support for org files

TerseTears:
Hello and thank you for the great plugin. 

I needed an overview for long `org` files hence this pull request. In principle, this acts similarly to the markdown feature, showing headlines in the specified pane:

Screenshot_20220418_124154

This uses the [tree-sitter-org](https://github.com/milisims/tree-sitter-org) parser, which isn't yet part of the nvim-treesitter repo.

I also ran the tests script and got errors for `markdown` and `org` which seems to be an issue with the `nvim-treesitter` plugin not properly recognizing custom parsers. Adding the line `vim.cmd([[runtime! plugin/orgmode.vim]])` (i.e. using the [orgmode](https://github.com/nvim-orgmode/orgmode) plugin) to the [tests/init.lua](https://github.com/stevearc/aerial.nvim/blob/3c64c5b76f2f09e23b91fd7c1dc29a6e0b953e68/tests/init.lua) file fixed the issue for `org` and its respective test passed successfully. 

So the tests work correctly, I'm just not sure if I should modify the [tests/init.lua](https://github.com/stevearc/aerial.nvim/blob/3c64c5b76f2f09e23b91fd7c1dc29a6e0b953e68/tests/init.lua) file since once `tree-sitter-org` becomes part of `nvim-treesitter`, said file will work as is, but not presently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ automatically fetch symbols from treesitter.
 - julia
 - lua
 - markdown
+- org
 - php
 - python
 - rst
--- a/lua/aerial/backends/treesitter/language_kind_map.lua
+++ b/lua/aerial/backends/treesitter/language_kind_map.lua
@@ -64,6 +64,9 @@ return {
   markdown = {
     atx_heading = "Interface",
   },
+  org = {
+    section = "Interface",
+  },
   php = {
     anonymous_function_creation_expression = "Function",
     class_declaration = "Class",
--- a/queries/org/aerial.scm
+++ b/queries/org/aerial.scm
@@ -1,0 +1,3 @@
+(section
+  (headline
+    item: (item) @name)) @type
--- a/tests/init.lua
+++ b/tests/init.lua
@@ -6,6 +6,8 @@ vim.o.swapfile = false
 vim.bo.swapfile = false
 -- Neovim 0.5 doesn't have julia filetype detection
 vim.cmd([[autocmd BufRead,BufNewFile *.jl setfiletype julia]])
+-- Neovim below 0.7 doesn't have org filetype detection
+vim.cmd([[autocmd BufRead,BufNewFile *.org setfiletype org]])
 
 require("nvim-treesitter.configs").setup({
   ensure_installed = "all",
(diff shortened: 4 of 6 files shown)


stevearc:
Thanks for the PR! I think you should be able to get the tests working by just adding the snippet from the parser repo. You can guard it in case it gets added to nvim-treesitter in the future:
```lua
local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
if not parser_config.org then
  parser_config.org = {
    install_info = {
      url = 'https://github.com/milisims/tree-sitter-org',
      revision = 'main',
      files = { 'src/parser.c', 'src/scanner.cc' },
    },
    filetype = 'org',
  }
end
```
If you put that in `tests/init.lua` right before the call to `setup()`, I *think* it should work.

You also may want to rebase onto the most recent master. I pushed up a bunch of commits to fix the CI for nvim 0.7.0 and the recent nvim-treesitter changes.

TerseTears:
Thank you. I made the changes, but I still believe this will be problematic, since it wouldn't actually install the `tree-sitter-org` parser. That would require changing the below line in Github actions and including `TSInstall org` as well:

https://github.com/stevearc/aerial.nvim/blob/657ed80259a2c59151c7f7c0d33f572df6fe7b69/.github/workflows/tests.yml#L35

Yet for `TSInstall org` to be a recognized command, Neovim will need to be started with the mentioned parser snippet...

I personally think the code is modified too much this way in unnecessary places. I will see if the merging of the org parser can come soon and reopen the request then if that is okay.

stevearc:
Personally I think we could make this work with just a couple more tweaks. Perhaps by adding a `tests/setup.lua` file that configures the org parser and calls `TSInstallSync all` and then sharing that logic with `tests/init.lua`. Then we could use that from the Github action instead of manually passing in `-c "TSInstallSync all"`.

But! If you'd prefer to get it merged into nvim-treesitter first, more power to you.

TerseTears:
Thank you @stevearc, it's just that authors are usually not this cool with having their code modified :-)

Regardless, the https://github.com/nvim-treesitter#2842 is now merged and should fix the issue, so I'm reopening this PR if you do not mind. I've also taken the liberty to add `org` to the list of supported languages in the README.

stevearc:
Awesome, thanks for the PR!

PR #87, opened 2022-04-23
fix: set buffer name after it's created

xeluxee:
Set a name for Aerial buffer
This is useful for
- statuslines to display name
- plugins using buffer names (e.g. `nvim-web-devicons` to display an icon along buffer name)


stevearc:
Looks good. Thanks for the PR!

stevearc:
This caused errors in certain situations. The issue is that buffer names must be unique, but more than one aerial buffer can exist at a time. I have reverted this PR for now; if you still want to have readable names for aerial buffers we'll have to find a way around that issue.

PR #91, opened 2022-05-02
fix: use server_capabilities instead of resolved_capabilities

ouuan:
```
[LSP] Accessing client.resolved_capabilities is deprecated, update your plugins or configuration to access client.server_capabilities instead.The new key/value pairs in server_capabilities directly match those defined in the language server protocol
```

https://github.com/neovim/neovim/issues/14090#issuecomment-1113956767


stevearc:
LGTM, thanks!

Issue #93, opened 2022-05-03
"LSP client not attached" since pr #91


AnthonyK213:
**Describe the bug**
`AerialInfo` -> "LSP client not attached"
Since #91 

**System information**
- OS: linux
- Neovim version:0.8.0
- AerialInfo: [
Filetype: lua
Configured backends:
  lsp (not supported) [LSP client not attached]
  treesitter (supported) (attached)
Show symbols: Function, Method
]
- Aerial config:
```lua
local kbd_b = vim.api.nvim_buf_set_keymap
local ntst = { noremap = true, silent = true }

require('aerial').setup {
    backends = {
        ['_'] = { "lsp", "treesitter" },
        markdown = { 'markdown' }
    },
    close_behavior = 'close',
    manage_folds = false,
    filter_kind = {
        ['_'] = {
            "Class",
            "Constructor",
            "Enum",
            "Function",
            "Interface",
            "Module",
            "Method",
            "Struct",
        },
        lua = {
            "Function",
            "Method",
        },
    },
    highlight_closest = false,
    on_attach = function (bufnr)
        kbd_b(bufnr, 'n', '{',          '<Cmd>AerialPrev<CR>', ntst)
        kbd_b(bufnr, 'n', '}',          '<Cmd>AerialNext<CR>', ntst)
        kbd_b(bufnr, 'n', '[[',         '<Cmd>AerialPrevUp<CR>', ntst)
        kbd_b(bufnr, 'n', ']]',         '<Cmd>AerialNextUp<CR>', ntst)
        kbd_b(bufnr, 'n', '<leader>mv', '<Cmd>AerialToggle!<CR>', ntst)
        kbd_b(bufnr, 'n', '<leader>fa', '<Cmd>Telescope aerial<CR>', ntst)
    end
}

require('telescope').load_extension('aerial')
```
```lua
local custom_attach = function (client, bufnr)
    -- ...
    require('aerial').on_attach(client, bufnr)
end

```

**Additional context**
Revert to  `7c65ec6254ae6939837930b8bf31fa351c77b069`  then everything is ok.

PR #94, opened 2022-05-03
`documentSymbol` -> `documentSymbolProvider`

AnthonyK213:
#93


stevearc:
Ah, should have tested that last PR. Thanks for the fix!

PR #113, opened 2022-06-04
feat(queries): add teal support

barklan:
Thanks for great plugin! This PR adds [teal](https://github.com/teal-language/tl) (a typed dialect of Lua) language support. It's treesitter queries are analogous to Lua's.


stevearc:
Looks great! Thanks for adding tests!

PR #114, opened 2022-06-04
feat(queries): add basic yaml support

barklan:
This PR adds basic support for YAML mappings and sequences.

In aerial window it looks nice:

window

..but in Telescope picker it is not ideal and difficult to search large files:

tele
 
Seems it would be nice if `value` column contained full key path (e.g. `services.proxy.ports`, for example [yaml.nvim](https://github.com/cuducos/yaml.nvim) plugin does something like that), but I am not sure how to do that correctly or if it's in the scope of this plugin.

### DIFF
--- a/lua/aerial/backends/treesitter/language_kind_map.lua
+++ b/lua/aerial/backends/treesitter/language_kind_map.lua
@@ -122,4 +122,8 @@ return {
   vim = {
     function_definition = "Function",
   },
+  yaml = {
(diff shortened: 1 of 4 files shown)


stevearc:
Looks great, thanks for the PR!

I added a30aa7991689545f4ee2259c4f44466d15952672 to show the full path to the symbol in the telescope picker. I believe that will do what you want.

PR #119, opened 2022-06-26
close_on_select should not close aerial on preview

WieeRd:
When `close_on_select` is set to true,  
aerial gets closed even when trying to preview the symbol definition with `}`, `{`, or `p`.  

Not very useful when just browsing through code using aerial,  
as this makes it impossible to actually "preview" the symbol definition.  

I think this option would be much more useful if it only closes aerial on "jump".  
(`<CR>`, `<C-v>`, `<C-s>`)


stevearc:
Makes sense to me. Thanks for the PR!

PR #131, opened 2022-08-03
Add support for Shoulda Context

threedaymonk:
[Shoulda](https://github.com/thoughtbot/shoulda) provides a similar test structuring DSL to that of RSpec. I'm working on a venerable project that uses Shoulda extensively, and it really helps to be able to move around the test files in a structured way.

This PR:

- Extends the block methods with those [defined by Shoulda Context](https://github.com/thoughtbot/shoulda-context) (`context`, `should`, etc.)
- Adds support for the one-line syntax available in Shoulda (e.g. `should validate_presence_of(:first_name)`)
- Incidentally fixes and adds test coverage for RSpec blocks with an optional argument (`before :all do`).

And as the list of methods became absurdly long, I've broken it across three lines.


stevearc (changes requested):
Thanks for the PR! Would you be able to add some tests to [ruby_spec.lua](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/ruby_spec.lua)? It doesn't have to cover all the cases, maybe just a normal "context" -> "should" block, and then something to cover the inline case you mentioned.

threedaymonk:
Gladly! I didn't realise there were tests, my bad.

threedaymonk:
@stevearc Is there any documentation of what's needed to run the tests? It doesn't work out the box, and even if I wrangle the `runtimepath` in `minimal_init.lua` to add the Plenary and Treesitter plugins, the tests themselves all die with the same error:

```
...
tree-sitter-typescript-49e82b1bce36d6046df911901684cd66b5345d58/examples/
tree-sitter-typescript-49e82b1bce36d6046df911901684cd66b5345d58/examples/parser.ts
nvim-treesitter[tsx]: Error during tarball extraction.
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

stevearc:
The tests _should_ just require plenary and nvim-treesitter. You can see what I'm installing in the CI here: https://github.com/stevearc/aerial.nvim/blob/master/.github/workflows/install_nvim.sh

I'm also using the `run_tests.sh` script to run them, but it's more of a convenience wrapper.

If you do have a lot of trouble running locally, you can always fall back to the "make a change, commit, push" and let github actions run the tests for you.

threedaymonk:
That was a useful exercise, because it made me realise that `setup` and `teardown` were being duplicated (`setup setup`), although it took me a long time to understand that the reason they were duplicated, and `before`/`after` weren't, was that the latter were being treated as a special case in `extensions.lua`.

In the course of doing so, I discovered a bug with the existing implementation of `before` and `after`. These can take an optional argument:

```ruby
before :all do
end
```

but this case appeared as just `:all`. The changes I made to `extensions.lua` incidentally fix this, but I've altered the test to cover this case.

lua/aerial/backends/treesitter/extensions.lua
```diff
@@ -139,7 +139,7 @@ M.ruby = {
     local method = (utils.get_at_path(match, "method") or {}).node
     if method then
       local fn = get_node_text(method, bufnr) or "<parse error>"
-      if fn ~= "before" and fn ~= "after" then
+      if fn ~= item.name then
```
threedaymonk, line 142: When there is no argument, `@method` and `@name` are the same. This implementation is both simpler than listing every method, and fixes optional arguments in `before` and `after`.

stevearc (approved):
Great work! Thanks for adding the tests!

PR #135, opened 2022-08-06
Support for Java fields

magidc:
I have extended Treessiter queries for java to capture Java fields. I have also extended Java test case file


queries/java/aerial.scm
```diff
@@ -9,3 +9,9 @@
 
 (enum_declaration
   name: (identifier) @name) @type
+
+(field_declaration
+  type: [(integral_type) (type_identifier) (boolean_type)] @java_type
```
stevearc, line 14: nit: remove the `@java_type` capture group; I don't think it's used

tests/treesitter/java_spec.lua
```diff
@@ -41,15 +41,33 @@ describe("treesitter java", function()
...
+            end_col = 22,
+          },
+          {
+            kind = "Field",
+            name = "field_2",
+            level = 1,
+            lnum = 8,
+            col = 2,
+            end_lnum = 8,
+            end_col = 21,
```
stevearc, line 60: Looks like tests expect this to be 24

stevearc (changes requested):
Awesome, thanks for the PR! Couple small nits and it'll be good to merge

tests/treesitter/java_spec.lua
magidc: yep, it should be 24, my fault

stevearc:
Thanks again!

magidc:
Thank you for developing Aerial! That was my first time dealing with TS queries

Issue #139, opened 2022-08-17
Scala treesitter support


mpasa:
It would be nice to have basic Scala support for Aerial. I can take some time checking how this is implemented and send a PR for the base constructs myself.

**Language**: scala

Please provide a minimal file in your language that includes all of the language constructs that you would like to see listed by aerial. Here is an example for lua: [lua_test.lua](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/lua_test.lua). Examples for other languages are in the same directory.

```
object Object {
  def foo(x: Int): Int = {
    5
  }
}

trait Trait {
  def foo(x: Int): Unit
}

class C {
  def foo(x: Int): Unit = {
    ???
  }
  def bar(x: Int): Unit = {
    ???
  }
}
```

If it is not obvious how each of the language constructs should map to a [LSP SymbolKind](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#symbolKind), please specify which SymbolKind to use for each of them.

mpasa:
Sent a PR for basic support on https://github.com/stevearc/aerial.nvim/pull/140

stevearc:
Is there additional support that you need for Scala, or can I close out this issue?

PR #140, opened 2022-08-17
Add basic Scala treesitter support

mpasa:
Adding basic support for the Scala language, solving https://github.com/stevearc/aerial.nvim/issues/139. It can potentially be extended, as this version only supports traits, objects, classes, and functions.


stevearc:
Awesome, thanks for the PR and for adding tests!

PR #143, opened 2022-08-22
feat: return line and column number for each symbol in get_location API

### LINKED ISSUE #142: Feature Request: `get_location` include line and column for each symbol
Thank you for all your hard work on Aerial! It has been a great help in my set up. I'm currently working on a fairly complicated `winbar` set up that I would like to allow me to click on a symbol and go to its location in the file. I was browsing through the code and saw that each symbol for navigation stores their line number and column number. Would it be possible to get that information passed as well when I get the table of symbols from the `get_location` function? This would allow me to dynamically build functions to control where to navigate when I click a symbol in my `winbar`. I'm not sure if there is any other type of metadata that would be useful to provide access to when you call the `get_location` function to get the current context, but this is one that I think is very useful.

Thanks again for this plugin!

mehalter:
Resolves #142

### DIFF
--- a/lua/aerial/init.lua
+++ b/lua/aerial/init.lua
@@ -150,6 +150,8 @@ M.get_location = function(exact)
       kind = item.kind,
       icon = config.get_icon(0, item.kind),
       name = item.name,
+      lnum = item.selection_range and item.selection_range.lnum or item.lnum,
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM, thanks for the PR!

PR #147, opened 2022-08-29
Add Treesitter Protobuf support

mpasa:
Adding basic Protobuf support for messages and enums.


stevearc:
Awesome! Thanks for adding tests!

PR #148, opened 2022-09-04
added support for Ruby setter methods

datanoise:
Adds support for Ruby setter methods like:

```ruby
def host=(value)
  @host = value
end
```


stevearc (changes requested):
Awesome! Could you add this [to the ruby test file](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/ruby_test.rb) as well?

datanoise:
Done

stevearc:
Fantastic, thank you!

PR #154, opened 2022-09-28
Add support for properties and fields for C# in treesitter backend

fnix2:
Fields and properties support.


stevearc:
Awesome, thanks for the PR! Would you be able to add a an [example](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/csharp_test.cs) and [test](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/csharp_spec.lua)?

fnix2:
I started doing it, but got confused. Could you explain how set "end_col" field in csharp_spec.lua?

stevearc:
The way I usually do it is I add the example, guess at the numbers, then inspect the test results when they fail and double check that they make sense. If they do, I copy those values into the spec file

fnix2:
Done

fnix2:
Btw in my configuration neovim with ```--noplugins``` and ```-u minimal_init.lua``` couldn't invoke treesitter and plenary. I needed to add this lines to minimal_init.lua:
```
vim.cmd("call plug#begin()")
vim.cmd("Plug 'nvim-treesitter/nvim-treesitter'")
vim.cmd("Plug 'nvim-lua/plenary.nvim'")
vim.cmd("call plug#end()")
```

Maybe it will help someone.

stevearc:
Excellent, thank you!

PR #157, opened 2022-10-02
Add support for text highlight groups in lualine module

JeremyWells227:
Support symbol text highlight groups in lualine module.

### DIFF
--- a/lua/lualine/components/aerial.lua
+++ b/lua/lualine/components/aerial.lua
@@ -55,7 +55,7 @@ local default_options = {
 
 local function color_icon(symbol_kind, icon, colored)
   if colored then
-    return string.format("%%#%s#%s%%##", "Aerial" .. symbol_kind .. "Icon", icon.." ")
+    return string.format("%%#%s#%s", "LuaLineAerial" .. symbol_kind .. "Icon", icon)
   else
     return icon
   end
@@ -63,7 +63,7 @@ end
 
 local function color(symbol_kind, text, colored)
   if colored then
-    return string.format("%%#%s#%s", "Aerial" .. symbol_kind, text )
+    return string.format("%%#%s#%s", "LuaLineAerial" .. symbol_kind, text)
   else
     return text
   end
@@ -80,12 +80,11 @@ local function format_status(symbols, depth, separator, icons_enabled, colored)
   end
 
   for _, symbol in ipairs(symbols) do
+    local name = color(symbol.kind, symbol.name, colored)
     if icons_enabled then
       local icon = color_icon(symbol.kind, symbol.icon, colored)
-      local name = color(symbol.kind, symbol.name, colored)
-      table.insert(parts, string.format("%s%s", icon, name))
+      table.insert(parts, string.format("%s %s", icon, name))
     else
-      local name = color(symbol.kind, symbol.name, colored)
       table.insert(parts, name)
     end
   end
--- a/lua/lualine/components/aerial.lua
+++ b/lua/lualine/components/aerial.lua
@@ -55,9 +55,17 @@ local default_options = {
 
 local function color_icon(symbol_kind, icon, colored)
   if colored then
-    return string.format("%%#%s#%s%%##", "Aerial" .. symbol_kind .. "Icon", icon)
+    return string.format("%%#%s#%s%%##", "Aerial" .. symbol_kind .. "Icon", icon.." ")
   else
     return icon
+  end
+end
+
+local function color(symbol_kind, text, colored)
+  if colored then
+    return string.format("%%#%s#%s", "Aerial" .. symbol_kind, text )
+  else
+    return text
   end
 end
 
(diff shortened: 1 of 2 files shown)


lua/lualine/components/aerial.lua
```diff
@@ -55,12 +55,20 @@ local default_options = {
 
 local function color_icon(symbol_kind, icon, colored)
   if colored then
-    return string.format("%%#%s#%s%%##", "Aerial" .. symbol_kind .. "Icon", icon)
+    return string.format("%%#%s#%s%%##", "Aerial" .. symbol_kind .. "Icon", icon.." ")
   else
     return icon
```
stevearc, line 60: By adding the space here, but not on line 60, there won't be a space between the icon and the symbol name when color is disabled. I'd prefer to keep the spacing logic where it is, unless there's a reason to change it?

lua/lualine/components/aerial.lua
```diff
@@ -74,12 +82,13 @@ local function format_status(symbols, depth, separator, icons_enabled, colored)
   for _, symbol in ipairs(symbols) do
     if icons_enabled then
       local icon = color_icon(symbol.kind, symbol.icon, colored)
-      table.insert(parts, string.format("%s %s", icon, symbol.name))
+      local name = color(symbol.kind, symbol.name, colored)
```
stevearc, line 85: You could extract this to outside of the if-block so you don't have to repeat it on line 88

stevearc (changes requested):
Thanks for the PR! Left a couple notes, and also please format with StyLua 0.14.3

lua/lualine/components/aerial.lua
JeremyWells227: That is a good point.  I wanted to handle the case where the syntax highlighting has a background color.  I think it might be better to change the status line string from `"%%#%s#%s%%##"` to `"%%#%s#%s"`.

image.

I will revert the spacing change and instead update the status line string.

After revisiting this, I think it would also be helpful to create a new set of highlights beginning with 
`"LuaLineAerial" .. symbol_kind`  

that links to the original, so that you can independently control how the syntax appears in lualine and in the aerials window.  I'll update this PR with that change to illustrate.

stevearc:
Question about the highlight groups: do you expect people to want to actually customize the colors of various symbols when they are displayed in the status line? Or is this just so that people can adjust the background color if it doesn't fit with lualine? If it's the latter, there are some utilities you can use to ensure that the bg color is transparent. You can see how I'm using that in a different plugin here: https://github.com/stevearc/overseer.nvim/blob/901da1301fd60c92972173e9466a358cba5174fd/lua/lualine/components/overseer.lua#L93-L95

JeremyWells227:
It's possible that the colorscheme and lualine theme could conflict, and a person would want to adjust certain icons.  In my case, I wanted it to have the same background as the lualine component, so I copied the highlight groups and manually substituted values from my theme. 

https://github.com/JeremyWells227/nvim/blob/6f0fb428a4ef13af9fd2cf40f386dbc2fed25d53/lua/plugin-setup.lua#L54-L117

If I understand what the snippet you linked is doing, it's defining a new highlight group at runtime that matches the lualine component?  I'll give a shot at replicating that.

lua/lualine/components/aerial.lua
JeremyWells227: I think this can be resolved, implemented in next commit.

lua/lualine/components/aerial.lua
```diff
@@ -53,15 +55,56 @@ local default_options = {
...
-local function color_icon(symbol_kind, icon, colored)
+function M:get_highlight(name)
+  if vim.fn.hlexists(name) ~= 1 then
+    return {}
+  end
+  local hl_test = vim.api.nvim_get_hl_by_name(name, true)
+  -- Linked to NONE
+  if hl_test[true] == 6 then
+    return {}
+  end
```
stevearc, line 66: This just means it's an empty dictionary. From `:help lua-special-tbl`:

> `{[vim.type_idx]=vim.types.dictionary}` is converted to an empty dictionary

And indeed, `vim.type_idx == true` and `vim.types.dictionary == 6`.

However, in this case I think we can eliminate this function entirely and just use 
```lua
require('lualine.utils.utils').extract_highlight_colors(group, 'fg')
```

lua/lualine/components/aerial.lua
```diff
@@ -53,15 +55,56 @@ local default_options = {
...
+    end
+  end
+  return hl_test
+end
+
+function M:color_for_lualine()
+  self.highlight_groups = {}
+  for _, symbol_kind in ipairs(identifiers) do
+    hl = "Aerial" .. symbol_kind
+    hl_icon = "Aerial" .. symbol_kind .. "Icon"
```
stevearc, line 79: nit: needs `local`

lua/lualine/components/aerial.lua
```diff
@@ -53,15 +55,56 @@ local default_options = {
...
 
-local function format_status(symbols, depth, separator, icons_enabled, colored)
+function M:color(symbol_kind, text, colored)
+  if colored then
+    local hl = self:format_hl(self.highlight_groups[symbol_kind].text)
+    return string.format("%s%s", hl, text)
+  else
+    return text
+  end
+end
```
stevearc, line 105: These two functions could be merged, or even possibly removed entirely.

```lua
local name = symbol.name
if colored then
  name = self:format_hl(self.highlight_groups[symbol.kind].text) .. name
end

local icon = symbol.icon
if colored then
  icon = self:format_hl(self.highlight_groups[symbol.kind].icon) .. icon
end
```

stevearc (changes requested):
Approach looks good! Just some comments about cleaning things up.

Apologies for the long wait between reviews. I've been traveling and moving, but am now in a more permanent location and should be able to be more responsive.

lua/lualine/components/aerial.lua
JeremyWells227: Ah that makes sense.  I had tried to use this function before writing my own and was unsuccessful, but I must've had some other issue that was causing it to quietly fail.

stevearc (approved):
Awesome, thanks for the contribution!

PR #188, opened 2022-11-28
Suggestion: use metadata "kind" in query file instead of `language_kind_map.lua`

monaqa:
Thank you for creating such a great plugin!
I especially love the nvim-treesitter integration feature.
However, I have one issue. I use my own tree-sitter parsers and queries for some minor languages, and when I try to display its outline, I can't extend the outline just by editing the query file.
I read the source a little and found that it is because it uses `language_kind_map.lua` to determine the kind of items in the outline. Unless I edit the plugin itself, I cannot add outline kinds at will.

So I have a suggestion: why not specify the kind from the metadata of the query file with `"#set!"`?
Using `(#set! "kind" "kind_name")` syntax, the kind can be specified for each pattern, and in the future `language_kind_map.lua` will no longer be needed.

Currently I have edited the Lua and Rust query files to try it out. I'd appreciate your feedback.


stevearc:
Seems fine to me! I think that's a pretty reasonable refactor. I'm happy to merge this, or wait if you wanted to do a broader migration.

monaqa:
Thank you for the review!
I will do the same for the other languages. Please wait a moment.

monaqa:
I added "kind" metadata to the other query files, removed `language_kind_map.lua`, and confirmed that `./run_tests.sh` prints "Success".
Please check it again!

stevearc:
Looks good to me! Thanks for the PR!

PR #203, opened 2023-01-05
fix open_automatic not working

cshuaimin:
I set `open_automatic = true` but aerial window not shows when I open a supported buffer. 

`lua=require'aerial.window'.maybe_open_automatic()` shows `E5108: Error executing lua ...vim/site/pack/lazy/opt/aerial.nvim/lua/aerial/window.lua:279: attempt to call field 'open_automatic' (a nil value)`, and `lua=require'aerial.config'.open_automatic` is `nil`.

So maybe it's not initialized?

### DIFF
--- a/lua/aerial/init.lua
+++ b/lua/aerial/init.lua
@@ -306,7 +306,7 @@ M.setup = function(opts)
     create_autocmds()
   end
 
-  if initialized then
+  if not initialized then
     do_setup()
   end
 end


stevearc:
This is not the fix; that logic is intended to only run `do_setup()` if the setup has _already_ completed, and defer it if not.

I fiddled around a bit and I think I've found what's causing the issue. Can you give it another try, and if you still have problems can you post an issue with your full aerial config?

cshuaimin:
ea86cd2 fixed the issue, thanks!

PR #205, opened 2023-01-06
fix: escape % in lualine component

Isrothy:
image
The following code makes lualine component crash because the symbol name contains '%'.

After this commit it should work.
image


stevearc:
Awesome, thanks for the PR!

PR #210, opened 2023-01-27
Use better default icon for strings

kluen:
It seems the icon for strings is not an icon. It looks a bit out of place, especially when looking at longer YAML files.

I chose the same icon as for text. I feel this is adequate here.

Before:
Screenshot 2023-01-27 at 17 06 15

After:
Screenshot 2023-01-27 at 17 08 12

Btw. it seems odd to me that the values all contain an additional space. I kept it for consistency though.


stevearc:
LGTM!

The extra space is so that wide symbols have enough space to be rendered. Not all symbols are wide, but even the ones that aren't get the extra space so they are all aligned the same. For more information about the rendering of wide symbols, [the kitty FAQ](https://sw.kovidgoyal.net/kitty/faq/#some-special-symbols-are-rendered-small-truncated-in-kitty) is an excellent resource.

PR #213, opened 2023-02-04
fix Lualine customize example in README.md

NeumoNeumo:
Appreciate for your convenient plugin!
When testing your example code to customize lualine, I found that the original code of ``lualine_y`` leaded to a warning after restarting neovim. It says: 
```
lualine: There are some issues with your config. Run :LualineNotices for details
```
According to the hints in ``LualineNotices``, it turns out that another pair of curly braces is required. So I add these two braces for new comers like me in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,7 +511,7 @@ require("lualine").setup({
     lualine_x = { "aerial" },
 
     -- Or you can customize it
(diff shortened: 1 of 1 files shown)


stevearc:
Good catch, thanks for the PR!

PR #216, opened 2023-02-05
elixir tree-sitter queries: add constants

emmanueltouzery:
tree-sitter: add constants for elixir.

see:
https://elixir-lang.org/getting-started/module-attributes.html#as-constants

In an ideal world, i'd almost prefer for the constants to be displayed with @ in the GUI (so to display `Module.@constant` instead of `Module.constant` -- that's what LSP does, but I didn't manage to achieve that...

still this is much better than not displaying constants at all.


stevearc:
LGTM. If you want to prefix constants with an `@`, it's pretty easy to do so. You would add the logic into the special-case elixir handling in 
https://github.com/stevearc/aerial.nvim/blob/7f09d0d00360e491cb063ceb56aac9b2d6e5a911/lua/aerial/backends/treesitter/extensions.lua#L71-L87

Happy to merge as-is, or wait if you want to make that change.

emmanueltouzery:
I'll make the change. But maybe not today. Possibly only tomorrow evening my time. Thank you for the explanation!

emmanueltouzery:
I have now made the change to prepend @ for constants.

PR #217, opened 2023-02-06
typescript: don't display in-function items

### LINKED ISSUE #214: feature: more configurable symbol filters
so here's with typescript and typescript react... and using the LSP backend, which is i believe better for that language..
many times in typescript (and javascript), it's better to define functions as constants:
```javascript
const myFunction = () => ...
```
instead of
```javascript
function myFunction() { ...
```

So that means that to actually see these functions, we must configure aerial with filter_kind containing Constant, problem solved, right?
Well, no, because then we get all the constants defined in the source code.. which is not what we want. The workaround I found for that is to filter these to only those that have children. That at least removes all the really simple constants. But it's a heuristic only.

And then.. the LSP will register as functions all the lambdas that we have all over in functions... for instance `x.map(() ->).filter(() -> )` will register two functions. But it registers these either with a symbol name ending with " callback" or with a symbol name of "&lt;function&gt;".

So again we can filter these by hand.

All of this to explain my feature request, which is to enable to configure which symbols to hide through callbacks in the aerial configuration. Example:

```lua
      filter_function = {
        ['_'] = nil,
        typescript = function(symbol)
          local kind = get_symbol_kind_name(symbol.kind)
          if kind == 'Constant' then
            return #symbol.children > 0
          end
          if kind == 'Function' then
            return string.match(symbol.name, " callback$") == nil and symbol.name ~= "<function>"
          end
          return true
        end
      }
```
No filter function in general, this special function for typescript. And this assume that `filter_kind` filtering is applied before this one, but this one could handle both.

With something like that, these kind of advanced (but I think relatively common) use-cases can be covered...

emmanueltouzery:
fixes #214

This PR makes it that we don't take into account variables that are nested under functions for typescript. This fixes my bug #214 -- the items are not necessarily marked as functions, but it's ok as long as they're displayed. However I don't want to fill in my display with local variables, and I guess most people also don't.

So it seems to me this fixes my personal issue and should be useful for everyone.

### DIFF
--- a/lua/aerial/backends/treesitter/extensions.lua
+++ b/lua/aerial/backends/treesitter/extensions.lua
@@ -290,6 +290,15 @@ M.typescript = {
       local str = vim.treesitter.query.get_node_text(string, bufnr) or "<parse error>"
       item.name = fn .. " " .. str
     end
+
+    -- we don't want to display in-function items
+    local cur_parent = value_node and value_node:parent()
+    while cur_parent do
+      if cur_parent:type() == "arrow_function" or cur_parent:type() == "function_declaration" then
+        return false
+      end
+      cur_parent = cur_parent:parent()
+    end
   end,
 }
 
--- a/tests/treesitter/ts_spec.lua
+++ b/tests/treesitter/ts_spec.lua
@@ -92,7 +92,7 @@ describe("treesitter ts", function()
         level = 0,
         lnum = 18,
         col = 0,
-        end_lnum = 30,
+        end_lnum = 32,
         end_col = 2,
         children = {
           {
@@ -188,6 +188,15 @@ describe("treesitter ts", function()
             },
           },
         },
+      },
+      {
+        kind = "Function",
+        name = "fn_4",
+        level = 0,
+        lnum = 34,
+        col = 0,
+        end_lnum = 37,
+        end_col = 1,
       },
     })
   end)
--- a/tests/treesitter/ts_test.ts
+++ b/tests/treesitter/ts_test.ts
@@ -27,4 +27,11 @@ describe("UnitTest", () => {
   describe.each([])("Test Suite", () => {
     test.each([])("runs multiple times", () => {});
   });
+  const local_const_var = "value";
+  let local_let_var = "value";
 });
+
+function fn_4() {
+  const local_const_var = "value";
+  let local_let_var = "value";
+}


stevearc:
While I 100% agree with your preferences, I've seen several instance of people using aerial and viewing _all_ symbols. I don't quite understand the use case, but it's enough that I wouldn't want to enforce this specific constraint on everyone. I think this is going to require some sort of user-facing config to allow people to flexibly define their own filtering logic.

emmanueltouzery:
What about introducing more symbol types that what comes with LSP or whatever? Aerial already allows to filter symbol types. For instance "LocalVariable".

emmanueltouzery:
But clearly allowing the user to post-process (and even add) symbols through a custom function is the most powerful. For instance I have a case where I would register symbols from custom code, or at least obscure libraries, it wouldn't make sense to include that support in aerial itself.

emmanueltouzery:
what do you think if, in `aerial.nvim/lua/aerial/backends/treesitter/init.lua`, at the end of `fetch_symbols_sync`, we called a user-configurable callback, if the user so configured?

```diff
  ext.postprocess_symbols(bufnr, items)
+ items = user_optional_postprocess(bufnr,  items, syntax_tree:root(), lang)
  backends.set_symbols("treesitter", bufnr, items)
```

that would allow to support a bunch of non-standard scenarios and user-specific tuning, but do you think it's a good idea?

stevearc:
I think that this is a reasonable approach. I would make a couple of tweaks:
1) instead of passing in the root & lang as parameters, make a general "context" table as the third parameter and add those to it
2) move this logic into `backends.set_symbols` so it applies the same regardless of the backend. Refactor `backends.set_symbols` to require a context table (and to add the backend name to that table before calling the user callback)

This way the user callback will be available regardless of which backend is being used, and we'll have a slightly more flexible API if we need to add new values to the callback later on.

emmanueltouzery:
Great, I'll prepare a PR to that effect ASAP!

emmanueltouzery:
I'm sorry, I would have another question.. the list of items that the user would take as input and return would be of type `@type aerial.Symbol[]`. The problem is that in the case of treesitter, that type would not include the treesitter "match", or the "value_node" (in the typescript treesitter postprocess, you take in the match and get a value_node from it, I'm not sure what's the relation between both).

If we give to the user the aerial.Symbol without the treesitter match info, the user would have to fetch back the symbol from the syntax tree themselves (presumably possible without too much trouble with the line/col position info). But we do already have that information...

What do you think would be better.. for each item...
1. we give to the user a plain `aerial.Symbol`, if they want to get back metadata, they can do that by hand
2. we give an `aerial.Symbol` and some info, maybe in an extra field of the aerial.Symbol, let's say called `source_info`. In the case of treesitter, the source_info would be an object containing either the match or value_node (which one? or both?), and would be nil (or the empty {}) for other sources.

emmanueltouzery:
There is maybe also the question that currently the list we would give to the user is post-processed enough to have hierarchical parent -child info. Which may be a little annoying for the user to process and world also require from him/her too return a correct parent/child structure.
My thinking related to that is that I wouldn't change that.. it's not that bad to loop on that, and it gives some more control to the user (and minimises the changes to make now). But maybe you don't agree?

stevearc:
So the current extension system for the treesitter backend has two main hook points. There's one that gets called on every symbol as it's getting added, and there's one that runs on the overall results after all symbols are collected.
https://github.com/stevearc/aerial.nvim/blob/a467e9a06aa92db81812f4c2c26364877bbafb80/lua/aerial/backends/treesitter/extensions.lua#L19-L20
There's another function above it to handle reparenting items, but we can mostly ignore that.

My inclination would be for user-facing extensions to follow roughly the same format, but be a bit more generic-ified so as to support the other backends. I think that would do what you want. What do you think?

emmanueltouzery:
I think two functions is fine, it absolutely fits my purpose, allows me to do all that I want! So great about that.
In my case, I would have my custom `postprocess` to filter out symbols.
And a custom `postprocess_symbols` add my app-specific symbols -- in my case I wrote a separate treesitter query .scm file and I apply it to the syntax tree to get extra app-specific symbols to include.

I do have an issue with the second part, adding symbols. The trouble is that they appear at the end of the list, and due to the hierarchy it's not trivial to sort things in order.
That problem is likely mostly present with the treesitter use-case.

In the treesitter case, we have in `aerial.nvim/lua/aerial/backends/treesitter/init.lua` the `stack` variable and we use it to derive a hierarchy for items. Now my custom postprocess_symbols is called afterwards, and I'm getting a hierarchy. And it's not trivial to put my items in that hierarchy when I don't have the `stack` variable anymore.

Actually (and i really don't know much the treesitter query system), I'm wondering whether maybe we'd be better off with not two, but three entry points: the two you've mentioned, and another earlier one: the user could specify her own .scm treesitter query files, per file type, besides the aerial one.

So we'd change:
```lua
  for match in query.iter_group_results(bufnr, "aerial", syntax_tree:root(), lang) do
```
to rather fetch all the matches from both query files (the aerial one, and the user one), and we would sort the matches in file order before we build the hierarchy. That would sort the issue of inserting items in-order without too many headaches for the user.
The third callback, `postprocess_symbols`, would then be useful to filter out or modify symbols, but probably not to add new ones, at least not for the tressitter codepath.

The problem with this however, is that it's treesitter-specific. I don't have a clear vision of all the backends that aerial supports.. So I'm interested in your feedback on the whole idea. Thank you!

emmanueltouzery:
I also want to thank you for your patience with me!

stevearc:
If two functions works for you, great! We should probably clean up the names, though. I didn't give it much thought at the time, but if it's user-facing it would be better to call them something clearer, like "post_parse_symbol" and "post_add_all_symbols".

If you want to extend the symbols for a language using treesitter, you can drop a query file into `after/queries/<lang>/aerial.scm` and put `; extends` at the top. That will automatically cause it to be picked up and used by aerial during the parsing step, and then you won't have to worry about changing the order/nesting in the postprocess step.

emmanueltouzery:
superseded by #220

PR #220, opened 2023-02-15
symbol post process APIs

emmanueltouzery:
APIs to allow the user to post-process the symbols detected by aerial.

This is the PR we've been discussing in #217 and #214.

Works great for me.. I tested it only for treesitter, but the changes are not complex and all the tests pass, so it should be OK for all the use-cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,37 @@ require("aerial").setup({
   -- Run this command after jumping to a symbol (false will disable)
   post_jump_cmd = "normal! zz",
 
+  -- Invoked after each symbol is parsed, can be used to modify the parsed item,
+  -- or to filter it by returning false.
+  --
+  -- bufnr: a neovim buffer number
+  -- item: of type aerial.Symbol
+  -- ctx: a record containing the following fields:
+  --   * backend_name: treesitter, lsp, man...
+  --   * lang: info about the language
+  --   * symbols?: specific to the lsp backend
+  --   * symbol?: specific to the lsp backend
+  --   * syntax_tree?: specific to the treesitter backend
+  --   * match?: specific to the treesitter backend, TS query match
+  post_parse_symbol = function(bufnr, item, ctx)
+    return true
+  end,
+
+  -- Invoked after all symbols have been parsed and post-processed,
+  -- allows to modify the symbol structure before final display
+  --
+  -- bufnr: a neovim buffer number
+  -- items: a collection of aerial.Symbol items, organized in a tree,
+  --        with 'parent' and 'children' fields
+  -- ctx: a record containing the following fields:
+  --   * backend_name: treesitter, lsp, man...
+  --   * lang: info about the language
+  --   * symbols?: specific to the lsp backend
+  --   * syntax_tree?: specific to the treesitter backend
+  post_add_all_symbols = function(bufnr, items, ctx)
+    return items
+  end,
+
   -- When true, aerial will automatically close after jumping to a symbol
   close_on_select = false,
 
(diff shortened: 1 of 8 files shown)


lua/aerial/backends/init.lua
```diff
@@ -161,9 +161,13 @@ M.set_symbols = function(backend_name, bufnr, items)
   local util = require("aerial.util")
   local window = require("aerial.window")
 
+  if config.post_add_all_symbols then
+    items = config.post_add_all_symbols(bufnr, items, ctx)
```
stevearc, line 165: Would be nice to add a nil-check here with a friendly error message in case the user mutates the symbols in-place and forgets to return them from the function.

README.md
```diff
@@ -372,6 +372,37 @@ require("aerial").setup({
...
+  --        with 'parent' and 'children' fields
+  -- ctx: a record containing the following fields:
+  --   * backend_name: treesitter, lsp, man...
+  --   * lang: info about the language
+  --   * symbols?: specific to the lsp backend
+  --   * syntax_tree?: specific to the treesitter backend
+  post_add_all_symbols = function(bufnr, items, ctx)
+    return items
+  end,
+
```
stevearc, line 405: This section is actually auto-generated, so you'll have to add it to `config.lua` and it will get copied here.

lua/aerial/backends/lsp/callbacks.lua
```diff
@@ -88,7 +89,18 @@ local function process_symbols(symbols, bufnr, fix_start_col)
           if symbol.children then
             item.children = _process_symbols(symbol.children, item, {}, level + 1)
           end
-          table.insert(list, item)
+          if
+            config.post_parse_symbol == nil
```
stevearc, line 93: Because you're going to have to add the example to the `config.lua`, the default value for this is going to be a function. Then, because of how the options are merged, it will become impossible for it to ever be `nil`. It would be nice to still let the user disable it by setting it to `false`, so maybe replace this with `not config.post_parse_symbol`

lua/aerial/backends/man.lua
```diff
@@ -36,8 +37,17 @@ M.fetch_symbols_sync = function(bufnr)
         lnum = lnum,
         col = 0,
       }
-      last_header = item
-      table.insert(items, item)
+      if
+        config.post_parse_symbol == nil
```
stevearc, line 41: Same

lua/aerial/backends/man.lua
```diff
@@ -49,18 +59,27 @@ M.fetch_symbols_sync = function(bufnr)
         end_lnum = lnum,
         end_col = line:len(),
       }
-      if last_header then
-        last_header.children = last_header.children or {}
-        table.insert(last_header.children, item)
-      else
-        table.insert(items, item)
+      if
+        config.post_parse_symbol == nil
```
stevearc, line 63: same

lua/aerial/backends/markdown.lua
```diff
@@ -37,7 +38,16 @@ M.fetch_symbols_sync = function(bufnr)
         end
         table.insert(parent.children, item)
       else
-        table.insert(items, item)
+        if
+          config.post_parse_symbol == nil
```
stevearc, line 42: same

stevearc (changes requested):
Really good! Just a couple small nits to look at.

README.md
emmanueltouzery: How does it get copied, which command should I run? Because presumably I must run the script before committing, right?

stevearc: It actually [gets run in CI](https://github.com/stevearc/aerial.nvim/blob/master/.github/workflows/update-docs.yml), but if you want to run it locally you can do `python .github/main.py generate`. Note that you will probably have to `pip install pyparsing==3.0.9` before, and you may want to do both of those from inside a virtualenv (this is why I have a python layout in the [.envrc](https://github.com/stevearc/aerial.nvim/blob/master/.envrc))

emmanueltouzery:
I have made a followup commit now with the changes :+1: 
About the config.lua and the settings, I'm not 100%, I just applied the changes to config.lua too, I believe the script would produce the same outcome as it is now with this commit.

stevearc:
Awesome, thanks for the contribution!

PR #221, opened 2023-02-20
elixir: skip apidocs

emmanueltouzery:
following my previous change related to elixir, apidocs were included in the symbols, which I think is counter-productive, skipping them with this change.


lua/aerial/backends/treesitter/extensions.lua
```diff
@@ -88,6 +88,9 @@ M.elixir = {
         item.name = string.format("describe %s", item.name)
       end
     elseif item.kind == "Constant" then
+      if item.name == "doc" or item.name == "moduledoc" then
+        return false
+      end
```
stevearc, line 93: Instead of adding the filter here, would it make more sense to make it part of the treesitter query? It looks like you're already filtering out certain constants there.

https://github.com/stevearc/aerial.nvim/blob/faadebfd77f176bd8acfab8bc9decac4abba26b0/queries/elixir/aerial.scm#L48-L52

stevearc (changes requested):
One Q

stevearc:
LGTM, thanks!

PR #226, opened 2023-03-11
ability to display private symbols differently

### LINKED ISSUE #225: feature request: differentiate public and private functions
it would be great if aerial could differentiate between public and private functions. When you want a quick overview of a file, looking first at the public functions would really help to get a picture as to what the file is about.

Currently as far as I'm aware, aerial doesn't support this.

In terms of GUI, this could be displayed through a different icon for public/private fields or functions, but maybe it would even make more sense to have different highlight colors of the same icon (so the same function icon, but with a different color depending on whether the function is public or private).

I'm not sure about the implementation though, because `SymbolKind` doesn't differentiate...

(Edit: I'm interested in this at first for elixir, where the public/private status is trivial to get from tree-sitter: there are two keywords: `def` and `defp`)

emmanueltouzery:
fixes #225

Implemented only the elixir private case. I thought just changing the icon wasn't enough, I had to make the text italic too. Maybe the icon should be the same color by default, this sets the icon for private symbols to NONE. I thought personally I needed all the help I could get to quickly spot the different public/private, so NONE suits me.

### DIFF
--- a/lua/aerial/highlight.lua
+++ b/lua/aerial/highlight.lua
@@ -29,6 +29,7 @@ M.create_highlight_groups = function()
   -- The name of the symbol
   for _, symbol_kind in ipairs(M.identifiers) do
     link(string.format("Aerial%s", symbol_kind), "NONE")
+    vim.cmd(string.format("hi Aerial%sPrivate cterm=italic gui=italic", symbol_kind))
   end
 
   -- The icon displayed to the left of the symbol
@@ -43,6 +44,7 @@ M.create_highlight_groups = function()
   link("AerialFieldIcon", "Identifier")
   link("AerialFileIcon", "Identifier")
   link("AerialFunctionIcon", "Function")
+  link("AerialFunctionPrivateIcon", "NONE")
   link("AerialInterfaceIcon", "Type")
   link("AerialKeyIcon", "Identifier")
   link("AerialMethodIcon", "Function")
--- a/README.md
+++ b/README.md
@@ -591,8 +591,9 @@ require("lualine").setup({
 ## Highlight
 
 There are highlight groups created for each `SymbolKind`. There will be one for
-the name of the symbol (`Aerial<SymbolKind>`, and one for the icon
-(`Aerial<SymbolKind>Icon`). For example:
+the name of the symbol (`Aerial<SymbolKind>`), one for name and scope (`Aerial<SymbolKind>Private`),
+one for the icon (`Aerial<SymbolKind>Icon`), and one for the icon with scope (`Aerial<SymbolKind>PrivateIcon`).
+For example:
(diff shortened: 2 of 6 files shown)


emmanueltouzery:
image
image

screenshots on my setup on aerial's `elixir_test.exs`.

lua/aerial/highlight.lua
```diff
@@ -29,6 +29,7 @@ M.create_highlight_groups = function()
   -- The name of the symbol
   for _, symbol_kind in ipairs(M.identifiers) do
     link(string.format("Aerial%s", symbol_kind), "NONE")
+    vim.cmd(string.format("hi Aerial%sPrivate cterm=italic gui=italic", symbol_kind))
```
stevearc, line 32: Would prefer for this to be the native lua function, and also to set the "default" highlight so it doesn't override settings that the user may already have in place
```lua
vim.api.nvim_set_hl(0, 'AerialPrivate', { default = true, italic = true})
```

stevearc (changes requested):
I think I would prefer adding another highlight group on top rather than creating another permutation for every SymbolKind. A function would have the hl group `AerialFunction`, a private function would have `AerialFunction` and then `AerialPrivate`. I think it provides better default behavior and scales better, but the downside is that if you want to have a different color for private functions vs private variables, it wouldn't support that. I think that's unlikely enough that the tradeoff is worth it. Thoughts?

In this world, I think we would still keep a separate group for the icons (e.g. `AerialFunctionIcon` + `AerialPrivateIcon`)

Could you also add `scope` to the [`aerial.Symbol` type definition](https://github.com/stevearc/aerial.nvim/blob/5b788392ec571621891e1b73887af5ac12056610/lua/aerial/data.lua#L10-L16)?

emmanueltouzery:
One option requiring only a single highlight group would be to have everything exactly as it is today, plus a new option `specific_highlight_groups_for_scopes`.. It could be set to for instance `["Private"]`.
If you set it to `[]` you have the exact behavior of today.
If you set it to `["Private"]` then a private function would not anymore get the `AerialFunction` highlight group, but `AerialPrivate`. And so on.

It is less flexible than what you suggested though. But it might still be enough.

More flexible would be a map...
`specific_highlight_groups_for = {Function = ["Private"]}`.

And so on. Something like that could work with only single highlight groups. Just an idea.

stevearc:
Check out #228 for a proposal. This is similar to the last option you mention (single group for private function/variable/type/etc), but also exposes a `config.get_highlight` configuration option that would allow users to change the highlight group dynamically for any particular combination of SymbolKind + scope. What do you think?

emmanueltouzery:
superseded by #228 plus #229

PR #228, opened 2023-03-16
refactor: change hl group calculation to be more customizable

stevearc:
To assist with #226

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,11 @@ require("aerial").setup({
     whitespace = "  ",
   },
 
+  -- Set this function to override the highlight groups for certain symbols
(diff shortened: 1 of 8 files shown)


emmanueltouzery:
maybe i would change this PR in a minor way, which would be to add a third parameter to `get_highlight`, which would be the default highlight name that aerial would use. That way the user could override the highlighting only in specific cases.. For instance in my case, I could say "if the scope is private return XX otherwise stick to the default aerial behavior".

stevearc:
@emmanueltouzery if you return `nil` it will use the default value. Is that sufficient?

emmanueltouzery:
Aha yes, absolutely, I missed that 👌

Issue #264, opened 2023-06-06
bug: aerial.nvim cannot complete cloning from a mirror site


gongfarmer:
### Neovim version (nvim -v)

v0.10.0-dev-309+g4e5061dba

### Operating system/version

MacOS 13.4, Debian 11, others

### Output of :AerialInfo

not available -- Aerial cannot complete installation

### Describe the bug

I am running neovim at a large company within a secure development environment.
There is no access to https://github.com/, however there is an internal mirror site with a different URL that provides access to GitHub projects.

The lazy.nvim package manager  gets plugins from our local mirror with this neovim configuration:

```lua
local opts = {
  git = {
    url_format = 'https://repomirror.corporate.com/github-neovim/%s.git',
  }
}
require('lazy').setup('plugins', opts)
```

This works for all the plugins we use except for aerial.nvim.
This plugin defines a submodule with a GitHub URL that is not reachable for us (in `.gitmodules`:

```
[submodule ".github/nvim_doc_tools"]
	path = .github/nvim_doc_tools
	url = https://github.com/stevearc/nvim_doc_tools
```

This has an [easy solution](https://confluence.atlassian.com/bitbucketserverkb/cloning-submodule-fails-from-mirror-946044514.html): Use a relative path for the url in `.gitmodules`.

### Steps To Reproduce

1. Be behind a corporate firewall which restricts access to https://github.com
2. Have access to a proxy containing mirrors of aerial.nvim and ivim_doc_tools
3.  Clone with submodules just like lazy.nvim package manager: `git clone --filter=blob:none --recurse-submodules --origin=origin  --progress  https://repomirror.corporate.com/git/stevearc/aerial.nvim.git`
4. Observe clone failure:
```
$ git clone --filter=blob:none --recurse-submodules --origin=origin  --progress  https://repomirror.corporate.com/git/stevearc/aerial.nvim.git
Cloning into 'aerial.nvim'...
Fetching objects: 4805, done.
Submodule '.github/nvim_doc_tools' (https://github.com/stevearc/nvim_doc_tools) registered for path '.github/nvim_doc_tools'
Cloning into '/home/bob/aerial.nvim/.github/nvim_doc_tools'...
fatal: unable to access 'https://github.com/stevearc/nvim_doc_tools/': gnutls_handshake() failed: Error in the pull function.
[...]

PR #265, opened 2023-06-06
fix: use relative url for submodule definition (#264)

gongfarmer:
Uses relative url for submodule so to enable users to clone from mirror site with `--recurse-submodules`.
Tested in  a site without GitHub access, using a mirror site
Tested with a normal configuration, using https://github.com as usual

Uses 2 levels of indirection (../../stevearc/nvim_doc_tools) instead of one (../nvim_doc_tools) so that people who fork aerial are not also forced to fork nvim_doc_tools.


stevearc:
Thanks for the PR!

PR #271, opened 2023-06-19
ci: add release-please to automate semantic versioning releases

mehalter:
This adds the release please github action for making release PRs to automate the release process and add semantic versioning tracking to Aerial! It automatically calculates the next version based on conventional commit messages which you are following :)

## Context

It is nice to have semantic versioning so users who use a plugin manager that can utilize that information can be let know when there are breaking changes easily and also set up the plugin managers to not update to the next major version to not accept breaking changes until they have time to handle the changes.

## Description

This adds the [release-please](https://github.com/google-github-actions/release-please-action) GitHub action which automatically calculates version numbers based on conventional commit messages and creates release PRs that can then be easily merged in when you are ready to tag the next release. It also handles managing changelogs and release notes in those releases as well.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,3 +56,27 @@ jobs:
       - name: Run tests
         run: |
           bash ./run_tests.sh
+
+  release:
+    name: release
+
+    if: ${{ github.ref == 'refs/heads/master' }}
+    needs:
+      - luacheck
+      - stylua
+      - run_tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          release-type: simple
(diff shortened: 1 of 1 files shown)


stevearc:
I'll try this out for a while and see how it goes. If it seems to be working well then I'll merge https://github.com/stevearc/dressing.nvim/pull/100 as well and maybe add it to some of my other plugins

mehalter:
Awesome! Thanks so much! It seems to be working so far, it opened PR #272 which if you merge it will automatically tag the version release with the changelog, and then will also make a `stable` tag that will be maintained to always point to the latest release.

Basically every time there is a commit to `main` the `release-please` action will continue to update the currently open PR for the next release and when you are ready to do the next release you can simply merge it :)

PR #273, opened 2023-06-25
Add solidity treesitter support

Eric-Song-Nop:
I have made the query for [solidity](https://soliditylang.org) and wrote tests for it.


stevearc:
Looks great! Thanks for the contribution!

PR #276, opened 2023-07-01
feat(markdown): setext_heading

tomtomjhj:


stevearc (changes requested):
Since you're adding support for more heading types, could you add them to [the test file](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/markdown_test.md) and update the tests?

tomtomjhj:
Added tests, but now the custom backend test fails because it doesn't implement setext_heading.

stevearc:
You can split the custom backend test into another file. It was only there because the functionality was equivalent, but now that it's not we can move it.

tomtomjhj:
fixed

stevearc:
LGTM thanks!

PR #277, opened 2023-07-10
elixir: support parameterless functions

emmanueltouzery:
it turns out elixir allows to skip the brackets for parameterless functions: aerial didn't take such functions into account so far.

### DIFF
--- a/queries/elixir/aerial.scm
+++ b/queries/elixir/aerial.scm
@@ -20,6 +20,7 @@
   (arguments [
               (call target: (identifier) @name)
               (binary_operator left: (call target: (identifier) @name))
+              ((identifier) @name)
    ])
   (#set! "kind" "Function")
   ) @type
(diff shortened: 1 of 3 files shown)


emmanueltouzery:
note that the tests failed due the julia query, unrelated to this PR. The elixir tests do pass.

queries/elixir/aerial.scm
```diff
@@ -24,6 +24,15 @@
...
 
+(call
+  target: (identifier) @identifier (#any-of? @identifier "def" "defp" "defguard" "defmacro" "defmacrop")
+  (arguments [
+              ((identifier) @name)
+   ])
+  (#set! "kind" "Function")
+  ) @type
+
+
```
stevearc, line 35: Is it possible to roll this into the above query by adding the `((identifier) @name)` to the alternation (`[]`)?

stevearc (changes requested):

queries/elixir/aerial.scm
emmanueltouzery: indeed you're right. had no idea [] is alternation in treesitter's syntax... this is what happens when one "learns by doing" instead of reading the docs :)

PR #283, opened 2023-07-30
improve icons

norflin321:


stevearc (changes requested):
Can you explain why you're making these changes and how you think they are better? On my terminal at least (kitty), it looks like most of the icons are the same, but more left-aligned instead of center-aligned. I actually prefer them to be more center-aligned because most of the other icons are center-aligned and it keeps them consistent.

norflin321:
Hello, original icons seem surprisingly small to me, new icons looks much bigger on my setup (macOS Ventura, kitty, nvim 0.10). I took them from another plugin that I use https://github.com/SmiteshP/nvim-gps/blob/master/lua/nvim-gps/init.lua.

Screenshot 2023-07-31 at 2 42 00 AM

Screenshot 2023-07-31 at 2 42 52 AM

stevearc:
Those icons are deprecated. If you search for the code points on https://www.nerdfonts.com/cheat-sheet you'll see that the ones you're adding are marked as "removed". For more information see https://github.com/ryanoasis/nerd-fonts/issues/1190#issuecomment-1530999114

I'm not sure why the existing icons aren't rendering properly for you; kitty should be the same even across operating systems

PR #287, opened 2023-09-04
feat(lualine): add separator highlight and beginning

kevmuko:
* Highlight the separator and enable an option `sep_highlight` to override the default `NonText`
* Add an option `sep_begin` to add the separator at the beginning for use with a preceding component

Preview:
Screenshot 2023-09-04 at 12 21 01 PM


stevearc:
LGTM thanks!

PR #289, opened 2023-09-14
fix: add JS / TS / TSX generator functions

madlabman:
Treat JavaScript generator expressions as functions declarations.


stevearc (changes requested):
Looks good! Can you add this to the `tsx` parser as well?
Also, it appears that the tests are not passing. You can run them locally with `make test`

madlabman:
> Looks good! Can you add this to the `tsx` parser as well? Also, it appears that the tests are not passing. You can run them locally with `make test`

Updated `tsx` as well, thank you!

I didn't figure out how to run tests previously, because I had tons of "ShaDa" parsing errors. Now the tests should pass.

PR #293, opened 2023-09-18
Expand ruby singleton handling

Slotos:
Ruby allows us to do a lot of weird stuff, and having weirdness glaring
at us from the side panel can be handy for when we sober up and start
cleaning.

This commit handles a number of singleton class manipulation cases,
as well as adds a reciever information to singleton method declarations.

image


stevearc:
LGTM, thanks!

PR #294, opened 2023-09-19
Replace parsers methods that are gone from next nvim-treesitter release

Slotos:
0.10 compatible nvim-treesitter is removing a lot of code, since main
  functionality is readily available in the core neovim.

This commit attempts to extract parsers calls that break functionality with nvim-treesitter's `main` branch.


Slotos:
Note that this is still insufficient to run tests with `main` nvim-treesitter branch.

lua/aerial/backends/util.lua
```diff
@@ -55,4 +55,59 @@ M.remove_change_watcher = function(bufnr, backend_name)
...
 end
 
+-- Taken directly out of nvim-treesitter with minor adjustments
+if (vim.treesitter.language == nil) or (vim.treesitter.language.get_lang == nil) then
+  ---@param bufnr nil|integer
+  M.get_buf_lang = function(bufnr)
+    local ok, parsers = pcall(require, "nvim-treesitter.parsers")
+    if not ok then
+      return false, "nvim-treesitter not found"
+    end
```
stevearc, line 65: This guard statement and the pcall can be removed, since if we get this far we can assume that nvim-treesitter is installed.

stevearc (changes requested):
Thanks for doing the migration! Left a couple of small requests

lua/aerial/backends/util.lua
```diff
@@ -55,4 +55,59 @@ M.remove_change_watcher = function(bufnr, backend_name)
...
+
+if (vim.treesitter.language == nil) or (vim.treesitter.language.get_filetypes == nil) then
+  ---@param lang string
+  M.has_parser = function(lang)
+    return require("nvim-treesitter.parsers").has_parser(lang)
+  end
+else
+  ---@param lang string
+  M.has_parser = function(lang)
+    if next(vim.treesitter.language.get_filetypes(lang)) ~= nil then
```
cshuaimin, line 92: In my test `get_filetypes()` always returns a non-empty table, even if the parser is not installed. I doubt if this function can be used.

lua/aerial/backends/treesitter/init.lua
```diff
@@ -15,12 +15,8 @@ local M = {}
 -- end (optional): The location of the end of this symbol (default @start)
 
 M.is_supported = function(bufnr)
-  local ok, parsers = pcall(require, "nvim-treesitter.parsers")
-  if not ok then
-    return false, "nvim-treesitter not found"
-  end
```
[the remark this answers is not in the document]
Slotos: Apologies for the long silence, finally getting back to this.

I'm realizing that in 0.10 (and probably in 0.9.1) we don't need nvim-treesitter itself. It's useful as a companion tool, but does not seem to be a requirement for aerial to function. Looks like every feature that is used is already present in neovim main.

I'm gonna make a pass with that assumption in mind.

Slotos:
This version pretty much eliminates reliance on nvim-treesitter helper code. It's not standalone (yet?), but code-wise everything relies on neovim own APIs.

There are four places nvim-treesitter code is directly called still:
- 0.8.3 fallback method (probably time to get rid of it, as 0.9.1 is now the minimal version for nvim-treesitter itself)
- two calls in test minimal config
- org spec, which could be switched to `aerial.backends.treesitter.util` if need be

-------

About minimal test config.

I removed query_predicates require, as it's not really required, but merely triggers some side-effect that stabilises nvim in test. I replaced it with a duplicate nvim-treesitter setup call, as it achieves the same outcome and will survive switch to `main`.

Removal of the setup call from minimal config results in two tests failing in `tests/attach_and_events_spec.lua`. I was able to reproduce this behaviour in interactive mode by additionally removing `require('nvim-treesitter').setup()` in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/plugin/nvim-treesitter.lua#L9.

Reproduction was performed with:
```bash
XDG_CONFIG_HOME=".testenv/config" \
      XDG_DATA_HOME=".testenv/data" \
      XDG_STATE_HOME=".testenv/state" \
      XDG_RUNTIME_DIR=".testenv/run" \
      XDG_CACHE_HOME=".testenv/cache" \
      nvim -u tests/minimal_init.lua
```

I was unable to figure out the reason behind such behaviour. `plugin/nvim-treesitter.lua` gets executed in test and interactive session alike, so there's probably some race condition that is being hit in test.

-------

`main` is still broken-ish, primarily in test. Test setup needs to be reworked, and there are a few additional races lurking in the shadows there, which I was unable to distill to an interactively reproducible case. On the contrary, in an interactive session aerial worked quite ok.

With `TSInstallSync` gone, the test harness will have to rely on install callbacks in order to wait for parsers installation. E.g.:

```bash
XDG_CONFIG_HOME=".testenv/config" \
  XDG_DATA_HOME=".testenv/data" \
  XDG_STATE_HOME=".testenv/state" \
  XDG_RUNTIME_DIR=".testenv/run" \
  XDG_CACHE_HOME=".testenv/cache" \
  nvim --headless -u tests/minimal_init.lua \
  -c "require('nvim-treesitter.install').install('all', { skip = { installed = true } }, function() require('plenary.test_harness').test_directory('${1-tests}', { minimal_init = './tests/minimal_init.lua' }) end)"
```

-------

Infodump over

lua/aerial/backends/util.lua
Slotos: Oh, you're absolutely right. I'd switched to using `get_string_parser` and recovering from error there. Even nvim-treesitter uses private APIs for now, so this might be a good enough solution until a better API becomes available.

Alternatively, `has_parser` and `get_buf_lang` can be rid of.
Starting with nvim 0.9, it's possible to simply grab a parser by buffer number, letting neovim figure it out.

Slotos:
Rebased on master, resolving conflict.

The latest commit changes matches processing in such a way that any value `#set!` in a query gets passed to aerial item.
As a practical example, I updated queries for ruby and elixir to set scope. In order to test it, I updated the test harness to also check "scope" field of the expectation.

Slotos:
Resolved conflicts with `master`.

stevearc (changes requested):
The migration looks good; I just added a couple of small tweaks.

The changes to the queries should be put into a different PR, as they don't have anything to do with treesitter version compatibility.

Slotos:
Applied comments, squashed history, deleted scope manipulation to be proposed in a separate PR.

lua/aerial/backends/treesitter/init.lua
```diff
@@ -68,7 +52,33 @@ M.fetch_symbols_sync = function(bufnr)
...
+    --- Matches can overlap. The last match wins.
+    local match = vim.tbl_extend("force", {}, metadata)
+    for id, node in pairs(matches) do
+      -- iter_group_results prefers `#set!` metadata, keeping the behaviour
+      match = vim.tbl_extend("keep", match, {
+        [query.captures[id]] = {
+          metadata = metadata[id],
+          node = node,
+        },
+      })
```
Slotos, line 79: This is the only extra change I applied, as I discovered that previously used method prefers metadata over captures, and even if this is mostly invisible, the behaviour should remain preserved until it needs to change.

stevearc:
Thanks for the PR!

PR #302, opened 2023-10-12
fix(render): use EOL extmarks to render AerialLine

folke:
I noticed that aerial highlight lines were not always correctly rendered. Especially when using edgy, that resizes component windows automatically.

Looking at the code, I found that to make a line, you used to pad with spaces, but this is not ideal.

Extmarks can be used instead that can extend to the EOL similar to the cursor line.

This PR implements that.

All the `highlight_modes` should work the same as before, but will now always properly extend till the end of the window when needed.

### DIFF
--- a/lua/aerial/render.lua
+++ b/lua/aerial/render.lua
@@ -15,7 +15,6 @@ end
 
 -- Resize all windows displaying this aerial buffer
 local function resize_all_wins(aer_bufnr, preferred_width, preferred_height)
-  local max_width = 0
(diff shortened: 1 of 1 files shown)


folke:
Thanks for looking into it! :)

Will do some work on neodev soon. Function defs have actually been created in core on nightly, so I will be able to remove some stuff from neodev. But also need to fix some things in core first. Soon!

stevearc:
Looks good, thanks!

PR #303, opened 2023-10-12
feat(lualine): added `sep_icon` option

folke:
Small PR to allow customizing the separator between the icon and the symbol.

Default separator is `" "` as before.

I personally prefer to set `sep = " "` and then `sep_icon = ""`.


stevearc:
LGTM!

folke:
Thanks!

PR #305, opened 2023-10-13
Feature: Set jumplist before jumping

### LINKED ISSUE #266: Feature Request: store locations in jumplist
Thanks for the excellent plugin! I would like to suggest a small tweak to better integrate with Neovim's existing navigation tools - saving the current location to the jumplist before jumping to a new symbol.

This would allow users to move between symbols with a combination of `aerial` and the standard `<c-o>`/`<c-i>`/etc. jump operators.

I think this should only take a 1-line (or so) change around here (although I haven't read through the codebase in depth): https://github.com/stevearc/aerial.nvim/blob/aff1bb8fecff83d3e3a2d544c4d4e6d65718bd19/lua/aerial/navigation.lua#L237-L241

Specifically, it should suffice to set the `'` register with `m'` just before jumping, if the user has jumping configured. If you're open to this, I'd be happy to make a PR.

wbthomason:
Closes #266. We may also want to add a configuration parameter if users should be able to disable/enable this behavior, but I think it's a reasonable enough default that it may be OK to wait for complaints before adding this.

I have not tested this extensively with weird window options/layouts, etc., but it has worked for my basic use case.

### DIFF
--- a/lua/aerial/navigation.lua
+++ b/lua/aerial/navigation.lua
@@ -237,6 +237,12 @@ M.select_symbol = function(item, winid, bufnr, opts)
     winid = vim.api.nvim_get_current_win()
     util.go_win_no_au(my_winid)
   end
+
+  if opts.jump then
+    local old_cursor = vim.api.nvim_win_get_cursor(winid)
+    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], old_cursor[1], old_cursor[2])
+  end
+
   vim.api.nvim_win_set_buf(winid, bufnr)
   local lnum = item.selection_range and item.selection_range.lnum or item.lnum
   local col = item.selection_range and item.selection_range.col or item.col
--- a/lua/aerial/navigation.lua
+++ b/lua/aerial/navigation.lua
@@ -240,7 +240,7 @@ M.select_symbol = function(item, winid, bufnr, opts)
 
   if opts.jump then
     local old_cursor = vim.api.nvim_win_get_cursor(winid)
-    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], old_cursor[1], old_cursor[2])
+    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], unpack(old_cursor))
   end
 
   vim.api.nvim_win_set_buf(winid, bufnr)


lua/aerial/navigation.lua
```diff
@@ -237,6 +237,12 @@ M.select_symbol = function(item, winid, bufnr, opts)
     winid = vim.api.nvim_get_current_win()
     util.go_win_no_au(my_winid)
   end
+
+  if opts.jump then
+    local old_cursor = vim.api.nvim_win_get_cursor(winid)
+    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], old_cursor[1], old_cursor[2])
```
max397574, line 243:
```suggestion
    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], unpack(old_cursor))
```

wbthomason: @max397574 Good suggestion, thanks!

lua/aerial/navigation.lua
```diff
@@ -237,6 +237,12 @@ M.select_symbol = function(item, winid, bufnr, opts)
     winid = vim.api.nvim_get_current_win()
     util.go_win_no_au(my_winid)
   end
+
+  if opts.jump then
+    local old_cursor = vim.api.nvim_win_get_cursor(winid)
+    vim.api.nvim_buf_set_mark(vim.api.nvim_get_current_buf(), [[']], unpack(old_cursor))
```
stevearc, line 243: I apologize for the terrible naming, this API needs some refactoring. The "jump" parameter doesn't mean what you think it means. When it's true, it means that we will focus the source window as part of this navigation action. So when you're just navigating around within the same buffer using `:AerialNext/Prev`, this value will be false.

Another thing to note: we're jumping to a location in `winid`, but `winid` is not necessarily the active window right here. You'll probably need to use `nvim_win_get_buf()` for the buffer, and you might need to do a `nvim_win_call` as well to make sure the mark is set in the appropriate window. If you want to test it, I would expect this to come up if your cursor is in the aerial window when you run `:AerialNext/Prev`.

stevearc (changes requested):
Added some notes about edge cases.

I will note that I did a couple of tests myself just playing around with setting the `'` mark, and had trouble getting it to behave correctly. I didn't look into it too much, but it's entirely possible that setting the marks this way via the API behaves differently than when you do it via the `m'` normal mode command.

yalikebaz:
Just wanted to nudge this a little, would be super helpful to have jumplist integrated here 🙏 
if not... I've customized my telescope picker to somewhat look like `AerialToggle`, and this works with the jumplist by default:

```
:lua require('telescope').extensions.aerial.aerial(require('telescope.themes').get_dropdown({layout_strategy='vertical', layout_config={vertical={anchor='E', width=0.2, height=0.9}}, previewer=false, sorting_strategy='ascending', results_title=false, prompt_title=false, border=false}))<CR>
```

stevearc:
Implemented by #424

PR #307, opened 2023-10-13
feat(telescope): only reverse results when `sorting_strategy = "descending"`

folke:
results in telescope are reversed to show them in order as they appear in the document.

However, this doesn't work properly when `defaults.sorting_strategy = "ascending"`.

In that case, the table should not be reversed.

# With this PR

image

# Without

`aerial_picker` should be on top

image


stevearc:
LGTM thanks!

folke:
Thanks!

PR #311, opened 2023-10-14
fix: delay when using `q` to close

chrisgrieser:
Same issue as with https://github.com/stevearc/dressing.nvim/pull/120


stevearc:
LGTM thanks

PR #316, opened 2023-10-16
Add snakemake

daler:
This adds some support for [Snakemake](https://snakemake.readthedocs.io/en/stable/). Since Snakemake is a superset of Python, this copies the existing Python `aerial.scm` and just adds the ability to navigate [rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html) via aerial.nvim.

I don't quite follow how to make the expected test cases in `test/symbols/`, so I copied over the Python one to see how the tests break to see if there are any clues...

### DIFF
--- a/queries/snakemake/aerial.scm
+++ b/queries/snakemake/aerial.scm
@@ -1,0 +1,19 @@
+(function_definition
+  name: (identifier) @name
+  (#set! "kind" "Function")
+  ) @type
+
+(class_definition
+  name: (identifier) @name
+  (#set! "kind" "Class")
+  ) @type
+
+(assignment
+  left: (_) @name
+  (#set! "kind" "Variable")
+  ) @type
+
+(rule_definition
+  name: (identifier) @name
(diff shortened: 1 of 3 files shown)


daler:
Ah, tests not enabled for PRs, it seems. @stevearc, `tests/symbols/snakemake_test.json` will definitely need to be fixed, I don't know how to create the expected results.

stevearc:
Tests are failing because there is no backend detected for the test file. This is because the filetype is not detected. You'll probably need to add another rule here:
https://github.com/stevearc/aerial.nvim/blob/34b820d6270fd149d4b3e1c5d697ee1dca365825/tests/minimal_init.lua#L9-L11

To update the symbols snapshot file, run `./run_tests.sh --update`. You'll need to have `jq` installed and in your PATH.

daler:
Thanks for the pointer -- tests passing locally now.

stevearc:
Thanks for the PR!

PR #317, opened 2023-10-17
Allow setting [privacy] scope from queries

Slotos:
This has direct impact on highlighting. If scope is set and set to anything but "public", highlighter will use a separate highlight for the corresponding entry. This was previously used by Elixir extension by manually processing nodes.

Elixir extension processing is removed, its queries updated, and fixed to consider not just `defp`, but also `defguardp` and `defmacrop` as private.

Ruby queries are updated to deal with both predicate and statement versions of scope setting. The query is kinda brittle still, as anything unexpected in between methods definitions will break it. "Not" match would solve the problem, but treesitter doesn't provide one. A custom predicate might be a way to go, but better is the enemy of good, so I'm stopping here.

### DIFF
--- a/lua/aerial/backends/treesitter/init.lua
+++ b/lua/aerial/backends/treesitter/init.lua
@@ -124,6 +124,7 @@ M.fetch_symbols_sync = function(bufnr)
       name = name,
       level = level,
       parent = parent_item,
+      scope = match.scope,
     }
     if use_selection_range then
       item.selection_range = selection_range
--- a/queries/ruby/aerial.scm
+++ b/queries/ruby/aerial.scm
@@ -4,20 +4,38 @@
   ) @type
 
 (method
-  name: (identifier) @name
+  name: (_) @name
   (#set! "kind" "Method")
   ) @type
 
-(method
-  name: (operator) @name
-  (#set! "kind" "Method")
-  ) @type
+(call
+  (identifier) @scope_switch
+  (#any-of? @scope_switch "private" "protected")
 
-(method
-  name: (setter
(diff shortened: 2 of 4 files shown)


queries/elixir/aerial.scm
```diff
@@ -25,6 +25,17 @@
...
+  target: (identifier) @identifier (#any-of? @identifier "defp" "defmacrop" "defguardp")
+  (arguments [
+              (call target: (identifier) @name)
+              (binary_operator left: (call target: (identifier) @name))
+              ((identifier) @name)
+   ])
+  (#set! "kind" "Function")
+  (#set! "scope" "private")
+  ) @type
+
```
stevearc, line 38: I think adding the scope with `#set!` is a good additional capability, but for the elixir case I feel like we should use the previous extension-based logic. It's so much cleaner to have a single if-statement to set the scope than to duplicate this entire query

stevearc (changes requested):
This looks good, apart from the note about the elixir changes.

You mentioned wanting to have a negated match; Neovim does have some built in predicates for that (see `:help lua-treesitter-not-predicate`). You might be able to use `#not-any-of?`

Slotos:
I ended up adding quite a few interleaving scope switches to ruby test case, since I'm relying on behaviours that I haven't found any documentation for.
Namely, the scoping query relies on specific nodes `(identifier)` winning against a catchall `(_)` one in an alternation query.

I have a few ideas about streamlining the whole proces by processing `@scope` node capture, but for now they are just that - ideas.

stevearc:
Cool, this LGTM!

PR #320, opened 2023-10-27
Added aerial queries for objdump

ColinKennedy:
Now that https://github.com/neovim/neovim/pull/23608 and https://github.com/nvim-treesitter/nvim-treesitter/commit/0179a89656b4ce395a4487c07ae385b8425524ae merged, objdump files are recognized in Neovim with tree-sitter support. Adding to aerial.nvim as well.

aerial nvim_objdump_demo

Since objdump files aren't really code, I wasn't sure what the preferred header would be. So I just chose `"Interface"` and `"Function"`. Open to suggestions, there.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,6 +72,13 @@ jobs:
         run: |
           bash ./run_tests.sh
 
+      - name: Check for changes
+        run: |
+          git add tests
+          git diff --quiet --staged || echo "Changes to snapshot file detected. Run 'make update_snapshots' and commit the changes"
+          git diff --stat --staged
+          git diff --quiet --staged
+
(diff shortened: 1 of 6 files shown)


stevearc (changes requested):
Interface & function sound fine to me; since the LSP symbol concepts don't really map to this use case it's somewhat arbitrary.

To make the tests pass you'll need to make sure the objdump file gets detected with the correct filetype, and you'll need to add a symbols snapshot file. See this comment for more info https://github.com/stevearc/aerial.nvim/pull/316#issuecomment-1765332801

ColinKennedy:
Thank you for the tip. It should be working now!

stevearc:
Thanks for the PR!

PR #323, opened 2023-11-08
Chores: Update markdown expectations, regenerate ruby symbols, enable nvim-treesitter main testing

Slotos:
# Markdown expectations

Recent update to nvim-treesitter breaks aerial.nvim tests.
Specifically:
- https://github.com/nvim-treesitter/nvim-treesitter/commit/96f55f30433
- https://github.com/MDeiml/tree-sitter-markdown/commit/cd705f15912

# Ruby symbols

Prior version held a bunch of data written by hand. This commit updates
it to avoid false alarms when generating symbols for new or updated
queries.

Expectations are not changed, only the order of keys in individual
tables was updated.

# Testing against nvim-treesitter main branch

While nvim-treesitter is not a strict requirement for aerial.nvim, it
is used to install parsers and it updates nvim default language to
parser assignments with practical values. This commit adds an ability to
run tests against `main` branch by providing `--test-main` option to
`run_tests.sh`.

### DIFF
--- a/tests/minimal_init.lua
+++ b/tests/minimal_init.lua
@@ -37,9 +37,22 @@ if master_nvim_ts then
     )
   end, { nargs = "?" })
 else
+  -- Use compiler that includes c++14 features by default
+  -- If `cc` doesn't implement those, override it for tests run with
+  -- `CC=gcc-13 ./run_tests.sh`
+  local parser_config = require("nvim-treesitter.parsers").configs
+  parser_config.norg = {
+    install_info = {
+      url = "https://github.com/nvim-neorg/tree-sitter-norg",
+      files = { "src/parser.c", "src/scanner.cc" },
+      branch = "main",
+    },
+    tier = 3,
+  }
+
   vim.api.nvim_create_user_command("RunTests", function(opts)
     local path = opts.fargs[1] or "tests"
-    require("nvim-treesitter.install").install("all", { skip = { installed = true } }, function()
+    require("nvim-treesitter.install").install(langs, { skip = { installed = true } }, function()
       vim.schedule(function()
         require("plenary.test_harness").test_directory(
           path,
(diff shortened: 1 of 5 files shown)


tests/minimal_init.lua
```diff
@@ -20,9 +20,39 @@ for lang, _ in vim.fs.dir("queries") do
...
+    local path = opts.fargs[1] or "tests"
+    require("plenary.test_harness").test_directory(
+      path,
+      { minimal_init = "./tests/minimal_init.lua" }
+    )
+  end, { nargs = "?" })
+else
+  vim.api.nvim_create_user_command("RunTests", function(opts)
+    local path = opts.fargs[1] or "tests"
+    require("nvim-treesitter.install").install("all", { skip = { installed = true } }, function()
```
stevearc, line 42: Does the new API support installing just the list of languages that we support?

stevearc (commented):
One Q

tests/minimal_init.lua
Slotos: Just took another look and it does. We can pass `langs` here.

As an aside, there's an issue with installing `norg` parser with `main` branch - it's not present there. If `master` is run first, the compiled parsers are visible to neovim, but if `main` is run on its own, an annoying issue arises.

Specifically, norg parser relies on c++14 standard. At the very least, default MacOS clang version does not enable required features by default, and fails to compile the parser. Providing `-std=c++13` results in `error: invalid argument '-std=c++14' not allowed with 'C'`.

Latest `clang` from homebrew works ok, and `gcc-13` also doesn't bat an aye.

In order to use a different compiler, `CC` environment variable can be used, e.g.:
```sh
CC=gcc-13 ./run_tests.sh --test-main
```

I just pushed a separate commit that adds neorg parser configuration and uses `langs` instead of `"all"` to install only the languages aerial handles.

stevearc:
Thanks for the PR!

PR #325, opened 2023-11-11
Cache aerial tree-sitter queries

Slotos:
When moving away from nvim-treesitter helpers, I overlooked the caching they were performing. This means that currently on every symbols refetch, neovim scans filesystem and gets all queries. Definitely useful when writing new queries, but slow nonetheless.

This commit introduces a simple query cache and an API method to clear said cache. With this, performance should improve, yet a way to iterate over query design remains accessible.

### DIFF
--- a/lua/aerial/init.lua
+++ b/lua/aerial/init.lua
@@ -459,6 +459,9 @@ M.nav_close = lazy("nav_view", "close")
 ---Toggle the nav windows open/closed
 M.nav_toggle = lazy("nav_view", "toggle")
 
+---Clear aerial's tree-sitter query cache
+M.treesitter_clear_query_cache = lazy("backends.treesitter.helpers", "clear_query_cache")
+
 ---Sync code folding with the current tree state.
 ---@param bufnr? integer
 ---@note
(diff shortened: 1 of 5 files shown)


lua/aerial/init.lua
```diff
@@ -113,6 +113,13 @@ local commands = {
       desc = "Close the aerial nav window.",
     },
   },
+  {
+    cmd = "AerialTSClear",
+    func = "treesitter_clear_query_cache",
+    defn = {
+      desc = "Clear aerial's tree-sitter query cache",
+    },
+  },
```
stevearc, line 122: It's probably quite uncommon for a user to need to do this, so I'd prefer to not add another command. There's already too many commands; I'd like to shift more of those to API-only functionality, the only thing stopping me is backwards-compatibility.

stevearc (changes requested):
One comment about the command

lua/aerial/init.lua
Slotos: Makes sense. I removed the command, leaving only an API method exposed.

stevearc (approved):
Awesome, thanks!

PR #332, opened 2023-12-07
vim treesitter: functions with "field expressions"

emmanueltouzery:
i'm very very far from a vimL specialist.. but I encountered this in https://github.com/kristijanhusak/vim-dadbod-ui/blob/master/autoload/db_ui/query.vim for instance


stevearc:
LGTM thanks!

PR #339, opened 2024-01-14
fix: remove extra double quotes for fzf prompt

231tr0n:
Remove extra double quotes around the fzf prompt which seems to be unecessary.
Without this fix merged the fzf prompt looks like this
image
With this fix merged the fzf prompt will look something like this without double quotes
image


stevearc:
LGTM

PR #350, opened 2024-03-11
[DRAFT] groovy support

emmanueltouzery:
some work on groovy support. this works for me, so my intention was to collect the JSON as produced by the test and store this as reference for the test, however it seems in the tests tree-sitter cannot parse my scheme file. I'm assuming it's because I'm using neovim 0.9 and probably a slightly older tree-sitter, while the tests run the latest versions of everything.

i'm hoping to migrate to 0.10 and upgrade my dependencies when 0.10 gets released, which should be soon (released planned on april first). At that point I could upgrade that PR.

however I thought I'd just dump this here for now, also in case I'm maybe mistaken and the reason for the test failures is somewhere else.

### DIFF
--- a/lua/aerial/backends/treesitter/init.lua
+++ b/lua/aerial/backends/treesitter/init.lua
@@ -19,6 +19,9 @@ M.is_supported = function(bufnr)
   if not helpers.has_parser(lang) then
     return false, string.format("No treesitter parser for %s", lang)
   end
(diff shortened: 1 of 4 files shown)


emmanueltouzery:
alright, it's not about 0.10, it's about 0.9 because my tree-sitter was outdated, and also i won't want to support 0.8 and the grammar was completely different. i'll close for now and reopen when i'll be closer.

PR #356, opened 2024-03-27
check if bufdata.last_win != nil

linusboehm:
When opening nvim in diffview (`nvim -c DiffviewOpen`) and then toggeling aerial, aerial errors when trying to call `vim.api.nvim_win_is_valid(bufdata.last_win)` as bufdata.last_win == nil. This change adds a check and returns in case of a nil value.


stevearc:
LGTM thanks!

PR #359, opened 2024-04-05
Add basic support for navigating zig files.

g-cassie:
This is very basic level support but it's enough for me personally to get value out of this plugin while doing zig development.

I tried to have it differentiate between methods and functions but it would end up dropping structs which contain methods.

### DIFF
--- a/queries/zig/aerial.scm
+++ b/queries/zig/aerial.scm
@@ -9,8 +9,24 @@
             ))))
     )
 
+(Decl
+    (VarDecl
+      variable_type_function: (IDENTIFIER) @name
+      (#set! "kind" "Struct")
+      (_
+        (_
+          (ContainerDecl
+            (ContainerDeclType "union") @symbol
+            ))))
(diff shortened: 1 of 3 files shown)


g-cassie:
I've found a couple rough edges on this. Will continue to work on ironing it out but let me know if you are okay with merging it eventually.

stevearc:
I'd definitely be up for merging! Just add a test file to https://github.com/stevearc/aerial.nvim/tree/master/tests/treesitter with at least one instance of each symbol type that is supported and run `make update_snapshots` to generate the symbol snapshot.

stevearc (changes requested):
Sending back to you until tests are added

g-cassie:
Thanks - test added.  Pretty nice testing setup!

stevearc:
Awesome, looks great! The test failure looks unrelated; I just need to adapt to something that changed upstream. Thanks for the contribution!

PR #360, opened 2024-04-09
Fix icon spacing for Nerd fonts

toupeira:
First of all, thanks for the very lovely plugin! :heart: 

While setting up the Lualine component and trying to make it as compact as possible, I noticed that there are two spaces between the icon and function name (in Aerial windows too).

Turns out one space is hard-coded in the icon definitions, probably to account for some of the double-wide characters? 

I believe this isn't necessary if you're using a terminal emulator which handles these correctly, or if you use the single-width monospace variants (e.g. I use Hack Nerd Font Mono instead of Hack Nerd Font Regular): https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting#why-do-the-glyphs-look-small-squished-or-not-full-width

This change might break with some terminal emulators, let me know if you want me to add a warning to the CHANGELOG or something. At least on Tilix it looks correct with either font variant, using [these examples here](https://github.com/stevearc/aerial.nvim/blob/master/lua/lualine/components/aerial.lua#L44-L53) with full-width icons separated by a single space:

| Full-width / Regular font |Single-width / Monospace font |
|---|---
| image | image |

The second commit also fixes `sep_icon` for dense mode, which I noticed was ignored :grinning:


stevearc (changes requested):
There is a space following these icons because kitty (my primary terminal emulator) uses that to properly display double-width icons (see https://sw.kovidgoyal.net/kitty/faq/#some-special-symbols-are-rendered-small-truncated-in-kitty). I understand that some terminal emulators will display the icons properly without the trailing space, but I don't use those :)

If you want to customize the icon spacing, you can do this via the config can't you?

toupeira:
@stevearc hmm I checked Kitty and it looks the same though, the only difference is that it automatically switches to the single-width version if there is no space following the characters, but the plugin still adds one (in addition to the other one from the icon definition, which this PR removes).

Just to be clear, do you not see two spaces on your machine on `master`? I don't see them in the demo video from the README either, so maybe something else is different in my setup.

What Kitty version and font do you use? I'm on 0.33.1 and using Hack Nerd Font. Here's how it looks for me:

| |  master | this PR |
|--- | ---|---
| Single-width character| image | image
| Double-width character | image | image

stevearc:
This is what I see on master
Screenshot 2024-04-16 at 4 34 31 PM

And this is what I see when I use the single width character

Screenshot 2024-04-16 at 4 34 44 PM

Using kitty 0.33.0 and Ubuntu Mono Nerd Font

toupeira:
@stevearc ok that's almost the same then :grinning: I tried out UbuntuMono here too and saw the icons are a bit wider there than with Hack, so I guess it makes sense to keep the extra space.

I dropped the first commit from the PR, which now leaves only the fix for dense mode.

stevearc:
Great, this fix looks good, thanks!

PR #377, opened 2024-06-17
support loongdoc

cathaysia:
make aerial.nvim works with https://github.com/cathaysia/tree-sitter-loongdoc


stevearc:
LGTM thanks!

PR #380, opened 2024-06-23
feat: added a highlight group for the left-hand side of the Nav float…

reverendpaco:
… ui to indicate the parent of the current symbol

I like the Ranger-like UI, but the left-hand side does not indicate which node is the parent of the nodes in the middle pane.

I created a highlight group AerialNavParent and linked it to AerialLine

Here's a pic:
image

Note how `Mappings.n` is highlighted on the left


stevearc:
Thanks for the PR! I did some refactoring to use a full-line highlight style and also re-use the `AerialLineNC` highlight because it serves a very similar purpose.

reverendpaco:
great.  thanks for considering and accepting

PR #381, opened 2024-06-25
Remove deprecation warning in the nvim-0.5 branch

tasuki:
The `nvim-0.5` branch should not have the deprecation warning for neovim versions smaller than `nvim-0.8`. This warning is for the `master` branch which requires newer neovim.


stevearc:
LGTM thanks!

PR #383, opened 2024-07-07
feat: add support for `mini.icons`

mehalter:
This adds explicit support for [`mini.icons`](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-icons.md) for providing icons to oil. This generalizes the icon providing into a function that figures out which icon provider is available (preferring `mini.icons` over `lspkind`) and returning a common icon providing interface.

This also uses the `glyph` style in `mini.icons` to control the `nerd_font = "auto"` option since `mini.icons` has a built in style for not using nerd fonts and using plain text.

Lastly, this makes sure that the generalized icon provider function is lazily evaluated when it's necessary rather than on setup and caches the value so that it doesn't have to re-calculate the provider each time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ require("aerial").setup({
   -- icon when the tree is collapsed at that symbol, or "Collapsed" to specify a
   -- default collapsed icon. The default icon set is determined by the
   -- "nerd_font" option below.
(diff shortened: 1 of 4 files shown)


stevearc:
LGTM thanks!

PR #395, opened 2024-07-29
Allow nesting symbol to be used with telescope and fix docs

### LINKED ISSUE #373: feature request: show nesting indent in telescope like sidebar and fzf-lua
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

indent information should be reserved in telescope too.
fzf-lua:
截屏2024-05-21 08 01 50

Telescope with nesed=true
截屏2024-05-21 08 03 17

### Provide background

_No response_

### What is the significance of this feature?

nice to have

### Additional details

_No response_

jjshoots:
This PR adds a nesting symbol in telescope, and also makes the readme a bit clearer when using with telescope to prevent #373.

### Examples of different nesting symbols
image
image
image

Closes #373

### DIFF
--- a/lua/telescope/_extensions/aerial.lua
+++ b/lua/telescope/_extensions/aerial.lua
@@ -13,6 +13,7 @@ local ext_config = {
     yaml = true,
   },
   nesting_symbol = false, -- false or string
+  nesting_symbol_length = -1, -- -1 for calculate from `nesting_symbol`
 }
 
 local function aerial_picker(opts)
@@ -44,10 +45,9 @@ local function aerial_picker(opts)
     ext_config.show_nesting[filetype] or ext_config.show_nesting["_"]
   )
   local nesting_symbol = ext_config.nesting_symbol
-  local nesting_symbol_length = 0
-  if nesting_symbol then
-    -- need to handle multibyte characters
-    nesting_symbol_length = vim.fn.strwidth(nesting_symbol)
+  local nesting_symbol_length = ext_config.nesting_symbol_length
+  if nesting_symbol_length == -1 then
+    nesting_symbol_length = #nesting_symbol
   end
 
   local backend = backends.get()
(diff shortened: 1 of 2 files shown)


lua/telescope/_extensions/aerial.lua
```diff
@@ -40,11 +40,12 @@ local function aerial_picker(opts)
     end
   end
 
-  if show_nesting == nil then
-    show_nesting = ext_config.show_nesting["_"]
-  end
-  local backend = backends.get()
+  local show_nesting = (
+    ext_config.show_nesting[filetype] or ext_config.show_nesting["_"]
+  )
```
stevearc, line 45: This logic is not the same, as it will fall back to the `_` key if `show_nesting[filetype]` is false, not just if it's `nil`

stevearc (changes requested):
Change looks fine to me, just one comment about a bug and please fix the stylua formatting

lua/telescope/_extensions/aerial.lua
jjshoots: Thanks for the catch, thought I was being smart but turns out maybe not. Reverted the change now.

stevearc:
Looking at it a bit more, I'm starting to dislike the API that I provided previously (`show_nesting`) and the extension you're adding I think highlights its deficiencies. This boolean will work for your use case, but then what about when the next person comes along and wants a slight tweak to the separator character, or wants it to behave differently for different file types?

I've refactored it to instead expose a format function. I think this is easier to understand than the filetype map, and it's far more flexible. You'll be able to get your "nesting symbol" behavior with 
```lua
function(symbol_path, filetype)
  if #symbol_path == 1 then
    return symbol_path[1]
  else
    return "> " .. symbol_path[#symbol_path]
  end
end
```

What do you think?

jjshoots:
Amazing, this looks great! Thanks for the thinking cycles!

PR #396, opened 2024-08-07
feat(toml): treesitter support for TOML tables

WieeRd:
I wrote some queries for TOML.
Captures tables as "Class" and arrays of tables as "Enum".

```toml
[foo]
[foo.bar]
[[baz]]
[[baz]]
[[neo."vim"]]
[[neo."vim"]]
```
```
C foo
C foo.bar
E baz
E baz
E neo."vim"
E neo."vim"
```

Unfortunately, `[foo.bar]` is not actually a child of `[foo]` in the AST.

```scheme
(document
  (table
    (bare_key)) ; foo
  (table
    (dotted_key ; foo.bar
      (bare_key) ; foo
      (bare_key))) ; bar
  (table_array_element
    (bare_key)) ; baz
  (table_array_element ; baz
    (bare_key))
  (table_array_element
    (dotted_key ; neo."vim"
      (bare_key) ; neo
      (quoted_key))) ; "vim"
  (table_array_element
    (dotted_key ; neo."vim"
      (bare_key)
      (quoted_key))))
```

As a result, they are displayed as separate top level tables `foo` and `foo.bar`.
Would it be possible to neatly nest `bar` under `foo` using more complex predicates?
I am still learning how to write queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ In addition, you will need to have either Treesitter or a working LSP client. Yo
 - snakemake
 - solidity
 - teal
+- toml
 - tsx
 - typescript
 - usd
--- a/queries/toml/aerial.scm
+++ b/queries/toml/aerial.scm
@@ -1,0 +1,17 @@
+(table
+  [
+    (bare_key)
+    (dotted_key)
+    (quoted_key)
+  ] @name
+  (#set! "kind" "Class")
+) @symbol
+
+(table_array_element
+  [
+    (bare_key)
+    (dotted_key)
+    (quoted_key)
+  ] @name
+  (#set! "kind" "Enum")
+) @symbol
--- a/tests/symbols/toml_test.json
+++ b/tests/symbols/toml_test.json
@@ -1,0 +1,167 @@
+[
+  {
+    "col": 0,
+    "end_col": 0,
+    "end_lnum": 9,
+    "kind": "Class",
+    "level": 0,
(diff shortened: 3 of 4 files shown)


stevearc (changes requested):
I don't think that more complex queries will fix the nesting, but I have a couple of hooks where you can add code to fix things like that after the fact. There isn't any documentation for this right now, but you can define an extension with the `get_parent` function similar to [what we're doing with markdown here](https://github.com/stevearc/aerial.nvim/blob/e75a3df2c20b3a98c786f5e61587d74a7a6b61d6/lua/aerial/backends/treesitter/extensions.lua#L111-L124). You will likely need to read and understand the [treesitter parsing logic](https://github.com/stevearc/aerial.nvim/blob/e75a3df2c20b3a98c786f5e61587d74a7a6b61d6/lua/aerial/backends/treesitter/init.lua#L25-L165) to know how that extension is used in context.

Also, please add a test file to the [tests/treesitter/](https://github.com/stevearc/aerial.nvim/tree/master/tests/treesitter) directory and then run `make update_snapshots` to generate a symbols snapshot file for the test file.

WieeRd:
Test document and symbols have been added.
But I don't think I'll be able to learn and debug the extension anytime soon.
How about we merge this for now and have a separate issue opened for the nesting of subtables?

stevearc:
LGTM thanks!

PR #397, opened 2024-08-08
Fix warning when closing buffers with the LSP backend in place

### LINKED ISSUE #393: bug: No LSP client found that supports symbols when deleteing buffer
### Neovim version (nvim -v)

NVIM v0.10.1 (github release)

### Operating system/version

Arch Linux

### Output of :AerialInfo

Aerial Info                                                                                                                                                    
-----------                                                                                                                                                    
Filetype: lua                                                                                                                                                  
Configured backends:                                                                                                                                           
  lsp (supported) (attached)                                                                                                                                   

### Describe the bug

When using Neovim v0.10 I am getting an error message, "No LSP client found that supports symbols in buffer X" where X is the buffer that I have just deleted. It seems like this is only happening on Neovim v0.10 and it does not happen on Neovim v0.9.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

0. This sets up the Lua language server, so expects that `lua-language-server` is in the $PATH. If you want to test with another language server you will have to modify the `repro.lua` file to set it up and adjust the commands below.
1. `nvim -u repro.lua`, spin up the minimal reproducing configuration below
2. `:e repro.lua`, open up a lua file so that the language server attaches
3. `:e test.lua`, open up a new lua file so we have a second one
4. `:bdelete`, close our new lua file
5. See the notification at the bottom

### Expected Behavior

There should be no error notification.

### Minimal example file

N/A the `repro.lua` is enough

### Minimal init.lua

```Lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- install plugins
require("lazy.minit").repro({
  spec = {
    {
      "stevearc/aerial.nvim",
      config = function()
        require("aerial").setup({
          backends = { "lsp" }, -- prefer LSP
          on_attach = function(bufnr)
            -- this function purely just has to exist
          end,
        })
      end,
    },
    {
      build = ":TSUpdate",
      config = function()
          auto_install = true,
          highlight = { enable = true },
        })
      end,
    },
    {
      "neovim/nvim-lspconfig",
      config = function()
        require("lspconfig").lua_ls.setup({})
      end,
    },
  },
})
```

### Additional context

_No response_

mehalter:
Fixes #393

# The Issue

Before closing a buffer all language servers are detached from it. This cause the `LspDetach` event to occur before the buffer is deleted and therefore is still valid. Part of the autocommand here when a language server detaches is it refreshes the symbols which is great when it is because a language server crashes or stopped by the user, but in the case of a buffer getting deleted is not the operation we really want to happen. It leads to a notification that no language servers are attached that support symbols. This is expected since the buffer has not been deleted yet (but will be immediately after)

# The Solution

It seems like this warning in `fetch_symbols` may not be necessary at all. There is a similar warning in the `is_supported` method of the LSP backend which I think covers the cases where this will really come up and affect the user. I spent a while trying to figure out the best way to circumvent the message, but because there is not way to know from the `LspDetach` event that the buffer is about to be deleted there's no real way to figure this out.

Let me know if you have any other ideas regarding this or another way to identify if a buffer is about to be deleted!

# Additional Information

I also noticed an extra check for `bufnr` elsewhere in the code when it is checked a few lines prior. I simply removed that as well as part of this. If you want me to remove that commit to focus the PR I totally can :)

Thanks so much for your amazing plugins and overall contributions to the Neovim community 😄

### DIFF
--- a/lua/aerial/backends/init.lua
+++ b/lua/aerial/backends/init.lua
@@ -95,6 +95,9 @@ local function attach(bufnr, backend, name)
   if not backend or not name or name == existing_backend_name then
     return false
   end
+  if not bufnr or bufnr == 0 then
+    bufnr = vim.api.nvim_get_current_buf()
+  end
   set_backend(bufnr, name)
   backend.attach(bufnr)
   if existing_backend_name then
--- a/lua/aerial/backends/init.lua
+++ b/lua/aerial/backends/init.lua
@@ -95,9 +95,6 @@ local function attach(bufnr, backend, name)
   if not backend or not name or name == existing_backend_name then
     return false
   end
-  if not bufnr or bufnr == 0 then
-    bufnr = vim.api.nvim_get_current_buf()
-  end
   set_backend(bufnr, name)
   backend.attach(bufnr)
   if existing_backend_name then
--- a/lua/aerial/backends/lsp/init.lua
+++ b/lua/aerial/backends/lsp/init.lua
@@ -39,10 +39,6 @@ M.fetch_symbols = function(bufnr)
   local params = { textDocument = vim.lsp.util.make_text_document_params(bufnr) }
   local client = lsp_util.get_client(bufnr)
   if not client then
-    vim.notify(
-      string.format("No LSP client found that supports symbols in buffer %d", bufnr),
-      vim.log.levels.WARN
-    )
     return
   end
   local request_success =


stevearc:
Thanks for the PR! I think it's fine to remove the notification here. I believe that it is left over from a much, much earlier version that didn't support multiple backends. These days, I would expect that the LSP backend wouldn't even attach if there aren't any clients that support document symbols.

PR #401, opened 2024-08-21
feat(just): treesitter support for Just

unicell:
This change adds treesitter query to support `just` filetype.

Ref: https://github.com/casey/just

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ In addition, you will need to have either Treesitter or a working LSP client. Yo
 - javascript
 - json
 - julia
+- just
 - latex
 - lua
 - make
--- a/queries/just/aerial.scm
+++ b/queries/just/aerial.scm
@@ -1,0 +1,4 @@
+(recipe
+  (recipe_header
+    name: (identifier) @name)
+  (#set! "kind" "Interface")) @symbol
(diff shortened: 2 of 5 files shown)


stevearc:
Looks like the tests are failing in nvim 0.9 because it doesn't detect the filetype. You can probably fix this by adding a line here: https://github.com/stevearc/aerial.nvim/blob/8d2a6c2ffad271b10c91f2c6f6ecef1f9625a9b9/tests/minimal_init.lua#L12

unicell:
> Looks like the tests are failing in nvim 0.9 because it doesn't detect the filetype. You can probably fix this by adding a line here:
> 
> https://github.com/stevearc/aerial.nvim/blob/8d2a6c2ffad271b10c91f2c6f6ecef1f9625a9b9/tests/minimal_init.lua#L12

Interesting. It must be my local plugin played a role in file type detection. Let me see if I can make tests happy by changing the extension to `.just` in test, since just support both. Otherwise, I'll patch it in minimal_init.lua.

stevearc:
Well the test failed on Neovim 0.9, so it's likely that the detection was only added in 0.10
https://github.com/neovim/neovim/commit/1338140ee3ea9355a9cc756f77b8fc71fd27eda5
That was after 0.9.0, and I don't think it was cherry-picked into any of the minor 0.9 releases.

stevearc:
Thanks for the PR!

PR #402, opened 2024-08-21
feat(starlark): treesitter support for Starlark

unicell:
feat(starlark): treesitter support for Starlark

This is for generating symbol outline for Bazel BUILD files.


stevearc:
LGTM thanks!

PR #412, opened 2024-09-21
fix telescope highlights regression from #386

emmanueltouzery:
the highlighting for the symbol type and the symbol name were lost with that change.

the telescope picker can be configured to display both symbol type and name and the line from the file.

#386 added highlighting for the line in the second part of the picker. however, doing so, it clobbered highlighting for the symbol type and name, which I am using thanks to the configurable `get_highlight` function to color differently public and private symbols.

this change makes sure the highlights for the first two columns are not clobbered by the line preview.

note, i'm not 100% sure about the offsets. i think there may be unicode widths shenanigans going on. but after this commit it's much better for sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,6 +605,10 @@ The extension can be customized with the following options:
 require("telescope").setup({
   extensions = {
     aerial = {
+      -- Set the width of the first two columns (the second
+      -- is relevant only when show_columns is set to 'both')
+      col1_width = 4,
+      col2_width = 30,
       -- How to format the symbols
       format_symbol = function(symbol_path, filetype)
         if filetype == "json" or filetype == "yaml" then
--- a/lua/telescope/_extensions/aerial.lua
+++ b/lua/telescope/_extensions/aerial.lua
@@ -100,16 +100,10 @@ local function aerial_picker(opts)
     buf_highlights = collect_buf_highlights() -- collect buffer highlights only if needed
   end
 
-  local function highlights_for_row(row, offset, icon_hl, name_hl)
+  local function highlights_for_row(row, offset)
     offset = offset or 0
     local row_highlights = buf_highlights[row] or {}
     local highlights = {}
-    if icon_hl ~= nil then
-      highlights = {
-        { { 0, COL1_WIDTH }, icon_hl },
-        { { COL1_WIDTH + 1, COL1_WIDTH + COL2_WIDTH + 3 }, name_hl },
-      }
-    end
     for _, value in ipairs(row_highlights) do
       local start_col, end_col, hl_type = unpack(value)
       hl_type = hl_type:match("^[^.]+") -- strip subtypes after dot
@@ -145,7 +139,13 @@ local function aerial_picker(opts)
       if #entry.name > layout[2].width then
         offset = offset + 2 -- '...' symbol
       end
-      highlights = highlights_for_row(row, offset, icon_hl, name_hl)
+      local col1_len = COL1_WIDTH + icon:len() - vim.api.nvim_strwidth(icon)
+      local col2_len = COL2_WIDTH + entry.name:len() - vim.api.nvim_strwidth(entry.name)
+      highlights = {
+        { { 0, col1_len }, icon_hl },
+        { { col1_len + 1, col1_len + 1 + col2_len + 1 }, name_hl },
+      }
+      vim.list_extend(highlights, highlights_for_row(row, offset))
     end
 
     return displayer(columns), highlights
--- a/lua/telescope/_extensions/aerial.lua
+++ b/lua/telescope/_extensions/aerial.lua
@@ -4,10 +4,9 @@ local finders = require("telescope.finders")
 local pickers = require("telescope.pickers")
 local telescope = require("telescope")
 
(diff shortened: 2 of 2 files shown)


stevearc:
I made some changes to hopefully account for multibyte characters. Can you check to see if it still works for you?

emmanueltouzery:
i think something is still off:
image

i guess too much bytes are given for the first column, which finishes too late, and one space is taken into account for the space between columns?

or maybe it's even more messed up, something is off with the preview here too:
image

without your changes i think the issue was mostly with the preview.

emmanueltouzery:
this is my commit without your changes btw:
image


and:
image

stevearc:
Hmmm...what version of Neovim are you on? And can you provide a small file that repros the weird highlights? This is what I'm seeing:

Screenshot 2024-10-05 at 4 57 05 PM

(with the icon and symbol name highlights overridden for testing, so they look bad but not incorrect)

emmanueltouzery:
hm, i didn't expect that you'd have issues reproducing the issue...

I'm using neovim 0.10.1. I'm on the telescope master branch, i wasn't at the latest commit but I now tried the latest commit and it's the same.
I also tried resetting my aerial configuration and it doesn't change anything.

This file with ft=typescriptreact shows the issue for me:

```typescript
import React, { ReactChild } from 'react';

export default class BootstrapSelect extends React.Component<Props> {
  static defaultProps = {
    valueType: 'string'
  };
  selectElement: null | HTMLSelectElement;

  domValueToValue(val: string | string[]) {
    return val;
  }

  valueToDomValue(value: Props['value']) {
    let val: Props['value'] | null = value;
    return val;
  }

  dom() {
    return $(this.selectElement!) as any;
  }
}
```

emmanueltouzery:
i was also on a 3 weeks old nvim-treesitter, tried upgrading it to the latest, no change.

stevearc:
Hmmm...still can't repro. Can you try with this config? I did `nvim -u repro.lua test.tsx` and this is what I see.

image

```lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

vim.g.mapleader = " "

local plugins = {
  "folke/tokyonight.nvim",
  {
    "nvim-telescope/telescope.nvim",
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    keys = {
      {
        "<leader>fs",
        function()
          require("telescope").load_extension("aerial")
          vim.cmd("Telescope aerial")
        end,
      },
    },
    opts = {},
  },
  {
    config = function()
        auto_install = true,
      })
    end,
  },
  {
    "emmanueltouzery/aerial.nvim",
    branch = "telescope_highlights",
    dependencies = {
      "nvim-tree/nvim-web-devicons",
    },
    opts = {},
  },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
```

emmanueltouzery:
thank you for the help with troubleshooting, much appreciated!

i'll research this further, probably tomorrow. for now i can't make it work, it installs the tree-sitter parser but somehow aerial doesn't find it:
image

emmanueltouzery:
so I realized why I have the wrong offsets for the highlights.. As you guessed it was due to my config.
I didn't notice it because it was in the telescope part of my config, as opposed to the aerial part of my config... but i have this:

```lua
local entry_display = require("telescope.pickers.entry_display")
_G.aerial_displayer = entry_display.create({
    separator = " ",
    items = {
      { width = 2 },
      { width = 32 },
      { remaining = true },
    },
  })
  require('telescope').extensions.aerial.aerial({
    get_entry_text = aerial_elixir_get_entry_text,
    displayer = aerial_displayer,
    attach_mappings = telescope_center_mappings,
    sorter = sorter,
  })
```

in other words i customized the aerial telescope picker to have different column widths (4->2 and then 30->32). and that worked without issues until the syntax highlight was added.

Maybe it could be possible for aerial to allow the user to customize such values, but then again i don't want to make everything too complex in your code. So unless you see a reasonable way to achieve that, i guess you could just merge this PR to fix the highlights for the first two columns, and move on... If you do find a way to make this customizable, i would appreciate it... if you have an idea that'd suit you, you can explain to me, and i'll gladly implement it.

emmanueltouzery:
I have now pushed a third commit in the branch to allow the user to configure the telescope col1 and col2 widths, which completely fixes my issues.

stevearc:
Great! Looks good to me!

PR #417, opened 2024-10-16
java: don't start methods at annotations

emmanueltouzery:
in the java tree-sitter hierarchy, annotations (`@Annotation`) before methods are nested within the `method_declaration`. In general they are indented on the line before the method itself (and sometimes there are many annotations, one per line, before the method).

The way aerial was capturing methods, in the telescope preview for instance, you'd see the annotation line instead of the method declaration line, and you were also taken to the "wrong" line.

image


stevearc:
Thanks for the PR!

PR #424, opened 2024-11-03
Feature: Add to jumplist before jumping

yalikebaz:
This PR adds the ability to use the <C-o> and <C-i> bindings to jump back and forth after selecting something from the aerial menu. I've tested this locally and seems to work well.


stevearc:
Tested it out and works fine for me! Thanks for the PR!

PR #428, opened 2024-11-28
feat: treesitter support for swift

simpleliujl:


tests/treesitter/swift_test.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+    init<T, V>(_ sss: T, xxx: V) {
+
+    }
+}
+
+// TODO: Some stuff
+// FIXME: Some fix
+// MARK: Some mark
```
stevearc, line 79: Any reason these are included?

stevearc (approved):
Generally LGTM, one question

tests/treesitter/swift_test.swift
simpleliujl: These are not necessory, I will delete them.

stevearc:
Thanks for the PR!

PR #436, opened 2024-12-25
Add scope into php method query

romankr-dv:
It enables the private scope highlighting for private php methods, for example:
```php
class MyClass {
    private function myMethod() { }
}
```
Thanks!

### DIFF
--- a/queries/php/aerial.scm
+++ b/queries/php/aerial.scm
@@ -13,6 +13,7 @@
   (#set! "kind" "Class")) @symbol
 
 (method_declaration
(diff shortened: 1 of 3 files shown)


stevearc:
Can we add some tests to the [php test file](https://github.com/stevearc/aerial.nvim/blob/master/tests/treesitter/php_test.php)?

romankr-dv:
The test cases for private, protected methods are added. Does it look good?

stevearc:
Yep, thanks for the PR!

PR #437, opened 2025-01-01
feat(fish): add treesitter query

phanen:


stevearc (changes requested):
Could you add a test file to this directory https://github.com/stevearc/aerial.nvim/tree/master/tests/treesitter and update the snapshots with `make update_snapshots` and confirm that the symbols look correct?

stevearc:
Thanks for the PR!

PR #438, opened 2025-01-01
feat: add query for nushell

melMass:
This adds basic symbol support for nushell 

https://github.com/user-attachments/assets/ddc7eeb9-111b-4f0f-9810-58a8dda71459

I'm also curious about two things:
- Is there a way to provide a literal `@name` ? For instance I wanted to collect nushell's `export-env` that don't have name, (I tried `(#set! "name" "MyName")` but it still returns Anonymous):
```nushell
export-env {
    $env.STEVE = arc
}
```
```scm
(decl_export
  (#set! "kind" "Module")
) @symbol
```
- Is there a way to provide custom HL? I would for instance use it to differentiate `export def` and `export --env def`, the later is allowed to modify the caller's environment variables


stevearc (changes requested):
Could you add a test file to this directory https://github.com/stevearc/aerial.nvim/tree/master/tests/treesitter and update the snapshots with make update_snapshots and confirm that the symbols look correct?

> Is there a way to provide a literal `@name`?

No, but you can add all kinds of custom behavior via an extension in [extensions.lua](https://github.com/stevearc/aerial.nvim/blob/master/lua/aerial/backends/treesitter/extensions.lua). Would that serve your purpose?

> Is there a way to provide custom HL?

No, because all the highlight groups have to be defined and well known so that users can supply their own overrides. This isn't a concept that maps to other languages, so there's nothing in the current highlight groups that would support it.

melMass:
Nice! Yep I realized right after submitting the PR and by a quick look couldn't tell how it worked, I'll do it now!

I'll check extensions I wasn't aware of this mechanism, but like that it definitely seems like what I was after!

melMass:
I made the update, strangely locally: groovy, julia, lua, ruby, ruby_singleton & swift returned a different "snapshot"

melMass:
Yep playing around with extensions there is surely and nifty way to solve all the cases I mentioned but that's definitely better in a dedicated PR

stevearc:
> I used Interface for const exports, this isn't really a concept I could find in other languages.

Would `Constant` work better for this? See the available SymbolKinds [in the LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind)

> It would be great to be able to differentiate "private" (non exported def) and public defs

There are a couple of languages that support `@scope`. Check out how we're doing that for ruby.

To get that test to pass, it looks like you'll need to add a filetype rule to the test setup https://github.com/stevearc/aerial.nvim/blob/b886c4b22c7563f29c5699699b485d6425cefb53/tests/minimal_init.lua#L7-L15

melMass:
> Would Constant work better for this?

Yep I will change it, my logic was that `Constant` isn't part of the default symbols (and you probably don't want it in the tree of most lang)

> There are a couple of languages that support @scope. Check out how we're doing that for ruby.

Interesting! I will need more experience with queries (and live experiment with it) to grasp what's possible as like that I don't fully understand all the syntax used

stevearc:

PR #446, opened 2025-02-14
feat(enforce): add enforce queries

simonvic:


tests/treesitter/enforce_test.c
```diff
@@ -0,0 +1,13 @@
...
+	BAZ
+}
+typedef Foo Bar;
+class Foo {
+	int m_foo;
+	void Foo() { }
+	void foo() { }
+	void bar();
+}
+// vim: ft=enforce
```
simonvic, line 13: the Enforce language uses `.c` file extensions; is it okay to use modeline to set the filetype for the test?

stevearc: Yep, whatever works! We're testing the parser, not vim's filetype detection.

PR #450, opened 2025-02-25
feat: add queries for R

wurli:
Adds support for R. Only adds one query for function definitions, but this is really all that's needed since R is a highly functional language.

I also added a test file `r_test.R`, but I wasn't able to get tests to run properly. Is there more to this than sourcing `tests/minimal_init.lua` and then using `:RunTests`?

Thanks!

### DIFF
--- a/queries/r/aerial.scm
+++ b/queries/r/aerial.scm
@@ -1,0 +1,5 @@
+(binary_operator
+    lhs: ((identifier) @name)
+    ["<-" "="]
+    rhs: (function_definition)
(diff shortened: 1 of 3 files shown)


stevearc:
You're going to want to run `make update_snapshots` to generate the symbols snapshot for the test, after which `make test` will run the tests

wurli:
> You're going to want to run `make update_snapshots` to generate the symbols snapshot for the test, after which `make test` will run the tests

Thanks! `symbols/r_test.json` now added :)

stevearc:

PR #460, opened 2025-04-10
fix: correctly set `last` property in `snack.picker` integration

mehalter:
This modifies the `snacks.picker` integration to keep track of the last element of a parent and correctly sets the `last` field to `true`.

Before:

2025-04-10_15:29:19_screenshot

After:

2025-04-10_15:29:09_screenshot


stevearc:
Thanks for the PR!

Issue #459, opened 2025-04-04
bug: chapters in tex files


AllergicMushroom:
### Neovim version (nvim -v)

NVIM v0.11.0

### Operating system/version

Artix Linux

### Output of :AerialInfo

Aerial Info                                                                                                                                                                     
-----------                                                                                                                                                                     
Filetype: tex                                                                                                                                                                   
Configured backends:                                                                                                                                                            
  treesitter (supported)                                                                                                                                                        
  lsp (supported)                                                                                                                                                               
  markdown (not supported) [Filetype is not markdown]                                                                                                                           
  asciidoc (not supported) [Filetype is not asciidoc]                                                                                                                           
  man (not supported) [Filetype is not man]                                                                                                                                     
Show symbols: Class, Constructor, Enum, Function, Interface, Module, Method, Struct 

### Describe the bug

Chapters declared with \chapter{...} do not appear on the outline. section and subsection seem to be fine.

The chapters appear in Vimtex ToC though, so I'm unsure where this issue is coming from. Perhaps a slight improvement to aerial would be to display the LSP name in :AerialInfo.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. Start a .tex file with book documentclass.
2. Declare two chapters with one section each.
3. Open Aerial outline. Sections appear, chapters don't.

### Expected Behavior

I expect chapters to also appear in the outline, and sections be nested inside the chapters.

### Minimal example file

\documentclass{book}

\begin{document}
 
\chapter{Chapter 1}

\section{Section 1}

\chapter{Chapter 2}

\section{Section 2}

\end{document}

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
[...]

PR #461, opened 2025-04-14
fix(latex): update queries for latex treesitter parser (#459)

helcoume:
Issue #459 highlights that the LaTeX Tree-sitter parser does not currently output \chapter sections.

Only \section, \subsection, and \subsubsection are supported by the parser at the moment.
According to [Overleaf’s documentation on sectioning](https://www.overleaf.com/learn/latex/Sections_and_chapters#Document_sectioning), \part and \chapter are additional valid sectioning levels. While \paragraph and \subparagraph are also described, they do not seem relevant to include here, in my opinion.

Changes made to address this issue:

* Add supports for \part and \chapter sections

* Add tests for \part and \chapter and update latex snapshot accordingly

### DIFF
--- a/queries/latex/aerial.scm
+++ b/queries/latex/aerial.scm
@@ -1,3 +1,19 @@
+(part
+  text: [
+    (curly_group
+      (text) @name)
+    (_)
+  ] @name
+  (#set! "kind" "Method")) @symbol
+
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR!

Issue #456, opened 2025-03-28
bug: split vertical bug


stuckinsnow:
### Neovim version (nvim -v)

latest

### Operating system/version

Mint

### Output of :AerialInfo

E5108: Error executing lua: ...p/.local/share/nvim/lazy/aerial.nvim/lua/aerial/util.lua:68: Invalid window id: 1012

### Describe the bug

Error received on successful launch of a new window, with float mode.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

When opening a float (specifically a float; this doesn't happen in sidebar or nav), and opening a split window from a function, it comes up with the above error.

### Expected Behavior

Everything should behave as is, except without the error.

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/aerial.nvim",
    config = function()
      require("aerial").setup({
        -- add your aerial config here
      })
    end,
  },
  {
    "nvim-treesitter/nvim-treesitter",
    build = ":TSUpdate",
    config = function()
      require("nvim-treesitter.configs").setup({
        ensure_installed = { "c", "lua" },
        auto_install = true,
        highlight = { enable = true },
      })
    end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

stevearc:
Using the `repro.lua` listed above, I opened a lua file and did `:AerialToggle float`. It opened correctly and produced no error. Is there anything else needed to repro this issue?

stuckinsnow:
[...]

PR #462, opened 2025-04-15
fix: prevent Aerial float window from closing on split (#456)

helcoume:
# Fix issue 'bug: split vertical bug #456'

## Problem

Calling a split-based action (e.g. aerial.action.jump_vsplit()) from a floating Aerial window triggers an error.
This happens because floating windows in Aerial are automatically closed via an autocmd when the window loses focus.

In `aerial.navigation.select_symbol()`, the split command doesn't disable these autocmds, so the floating window gets closed during the split. When the code attempts to return to the now-closed window, it raises an error.

## Solution

Prefix the split command with `noautocmd` (`noau`) to prevent triggering the window-closing autocmd.
This aligns with the behavior of `aerial.util.go_win_no_au`, which also uses `noau` for safe window switching.

### DIFF
--- a/lua/aerial/navigation.lua
+++ b/lua/aerial/navigation.lua
@@ -233,7 +233,7 @@ M.select_symbol = function(item, winid, bufnr, opts)
     end
     local my_winid = vim.api.nvim_get_current_win()
     util.go_win_no_au(winid)
-    vim.cmd(split)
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM thanks!

PR #465, opened 2025-05-07
fix(telescope): correct highlight offset computation (#414)

helcoume:
# Fix issue #414: highlights sometimes off by 1 in Telescope extension

This PR fixes incorrect offset computation for Tree-sitter highlights in the Telescope extension, which could cause them to appear shifted by one (or more) character.

## Root Cause

The issue tries to compute the highlight offset before calling `displayer()`. This approach is error-prone due to the presence of multibyte characters, single-byte characters, and potential line truncation, which makes it difficult to reliably calculate visual alignment.

## Solution

The fix changes the order of operations: it now calls `displayer()` first to get the final rendered string, then calculates the highlight offset by searching for the start position of the text to highlight within that string.

## Additional Fix

The issue also noted that no highlights were applied when `show_columns` was set to "symbols". This PR enables highlights in that mode as well, since icons are displayed (and usually highlighted).

### DIFF
--- a/lua/telescope/_extensions/aerial.lua
+++ b/lua/telescope/_extensions/aerial.lua
@@ -128,26 +128,23 @@ local function aerial_picker(opts)
       { entry.name, name_hl },
     }
 
-    local highlights = {}
-    if show_columns == "both" then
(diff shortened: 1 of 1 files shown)


stevearc:
LGTM thanks!

PR #476, opened 2025-07-24
fix: the lualine component become empty after calling :edit.

adoyle-h:
Reason: `:edit` will trigger autocmd BufUnload and then data.delete_buf(bufnr). So the data has no symbols.

The content of lualine comes from get_status_normal() and get_status_dense() which formating symbols. But get_location return empty,

```
  if not data.has_symbols(0) then
    return {}
  end
```

So we need to fetch symbols on BufEnter.

### DIFF
--- a/lua/aerial/autocommands.lua
+++ b/lua/aerial/autocommands.lua
@@ -77,7 +77,7 @@ local function update_aerial_windows()
   end
 end
 
-M.on_enter_buffer = util.throttle(function()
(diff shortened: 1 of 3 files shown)


stevearc (changes requested):
As an alternative, what do you think about changing the `BufUnload` autocmd to trigger on `BufWipeout`? Does that fix your issue?

adoyle-h:
> BufUnload

Yes, that fixes my issue.

PR #481, opened 2025-08-23
fix(c): remove buggy query pattern

### LINKED ISSUE #413: C query issue: Function after global variable is not detected
**Language**: C
**AerialInfo**:
```
Aerial Info
-----------
Filetype: c
Configured backends:
  treesitter (supported) (attached)
  lsp (supported)
  markdown (not supported) [Filetype is not markdown]
  asciidoc (not supported) [Filetype is not asciidoc]
  man (not supported) [Filetype is not man]
```

The minimal amount of code that demonstrates the missing symbol:
```C
01  int var = 5;
02  
03  int foo() { // undetected
04      return 0;
05  }
06  
07  int bar() {
08      return 0;
09  }
```

**Expected symbol**: Function named `foo` at lines 3-5

**Explanation**

A function following a global variable declaration doesn't appear to be detected. It appears that lines from 1-5 are detected as being part of a function named `var`, but `var` is actually just a variable on line 1 (also happens with arrays). Here is the output from `get_location(true)` at line 4:

```
{
  {
    col = 4,
    icon = "󰊕 ",
    kind = "Function",
    lnum = 1,
    name = "var"
  }
}
```

The C++ query file doesn't seem to have this issue when identical code is used (putting it in a `.cpp` file instead of a `.c` file). Perhaps a solution is just to use the same query file for both languages (or a stripped down version of the C++ query file)? Actually it looks you can't simply do that. I guess the parsers are different. Unfortunately I don't know enough about treesitter queries to fix this myself.

Thanks in advance

rmuir:
Reported bug of missing functions (#413) is caused by this query pattern:

```
((declaration) @root @start
  .
  (function_definition) @symbol @end
  (#set! "kind" "Function"))
```

Removing the pattern fixes the bug and causes no regressions in tests. The pattern was introduced in https://github.com/stevearc/aerial.nvim/commit/91350456c176fe5ef72e342dd3a75f726805454d but it is not clear what its purpose is: all functions should be matched by the simple `(function_definition)` query immediately after it.

The example from #413 and a few additional tests from tree-sitter-c's test corpus were added.

Closes #413

Note: if by any chance you happen to remember what this pattern was trying to accomplish, I can look to address/test for it.

### DIFF
--- a/queries/c/aerial.scm
+++ b/queries/c/aerial.scm
@@ -8,10 +8,5 @@
   declarator: (type_identifier) @name
   (#set! "kind" "Struct")) @start
 
-((declaration) @root @start
-  .
-  (function_definition) @symbol @end
-  (#set! "kind" "Function"))
-
 (function_definition
   (#set! "kind" "Function")) @symbol @root
--- a/tests/symbols/c_test.json
+++ b/tests/symbols/c_test.json
@@ -103,5 +103,80 @@
       "end_lnum": 19,
       "lnum": 19
     }
+  },
+  {
+    "col": 0,
+    "end_col": 1,
+    "end_lnum": 25,
+    "kind": "Function",
+    "level": 0,
+    "lnum": 23,
+    "name": "foo",
+    "selection_range": {
+      "col": 4,
+      "end_col": 7,
+      "end_lnum": 23,
+      "lnum": 23
+    }
+  },
+  {
+    "col": 0,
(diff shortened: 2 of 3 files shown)


stevearc:
The tests should have 100% coverage of the treesitter queries, so you can do pretty much whatever with the queries as long as the tests still pass. Thanks for the fix!

PR #484, opened 2025-09-02
feat: treesitter queries for editorconfig, hcl, ssh_config

rmuir:
Terraform is a different treesitter parser than hcl, but for aerial purposes these are the same. Support hcl also, it is helpful for docker bake files at least.

ssh_config / editorconfig are a few more "devops" formats with simple "sections", aerial helps when the files are large. The editorconfig pattern is a bit strange, but this is due to the grammar: any fancy stuff such as `not-kind-eq` is avoided since this predicate isn't in neovim core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,13 @@ In addition, you will need to have either Treesitter or a working LSP client. Yo
 - css
 - dart
 - djot
+- editorconfig
 - elixir
 - enforce
 - fish
 - go
 - groovy
+- hcl
 - help
 - html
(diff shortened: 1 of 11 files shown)


rmuir:
This PR may seem a bit crazy, as you might normally consider some of these files to be tiny... but I've got > ~1K lines for all of them. For example, java editorconfigs with all the intellij formatting cases get big: https://github.com/apache/lucene/blob/main/.editorconfig

PR #485, opened 2025-09-02
feat(yaml): treesitter name array elements symbol based upon "name" key

rmuir:
Previously, these elements were not listed as symbols. redhat's yaml-language-server lists these as numbered elements such as 0, 1, 2, but treesitter strategy didn't match them at all.

There's a de-facto standard for many yaml formats, to use the "name" key to solve this problem. It is a best practice for commonly used yaml files such as github workflows and ansible.

Note: I can tune the query if there are concerns: I'm aware this looks a bit strange, but it really makes these files possible to navigate, and doesn't hurt anything else (only introduces symbol where one didn't exist before). Try it out on some actions/ansible files.

Screen_Shot_2025-09-02_at_02 14 57

### DIFF
--- a/queries/yaml/aerial.scm
+++ b/queries/yaml/aerial.scm
@@ -9,3 +9,19 @@
   value: (block_node
     (block_sequence) @symbol)
   (#set! "kind" "Enum")) @start
+
+; "array element" with a "name:" key
+; ansible play/task names, etc
+; github workflow/actions jobs/steps, etc
+(block_sequence_item
+  (block_node
+    (block_mapping
+      (block_mapping_pair
+        key: (flow_node
+          (plain_scalar
+            (string_scalar) @_key))
+        value: (flow_node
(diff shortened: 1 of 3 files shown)


stevearc:
Yeah this seems like a good improvement, and is in line with some of the other specific cases made for other languages (e.g. rspec tests). Thanks!

rmuir:
@ZachVec personally I use treesitter, falling back to LSP (which I believe is the default with this plugin). 

There are a few exceptions where I prefer LSP. Hope this helps:
```lua
  backends = {
    -- jdtls does a fantastic job
    java = { 'lsp', 'treesitter' },
    -- gopls does better for now
    go = { 'lsp', 'treesitter' },
    -- lua: don't fall back to lsp, no symbols is better than crazy symbols (emmylua_ls)
    lua = { 'treesitter' },
  },
```

PR #494, opened 2025-11-01
fix: lspkind symbol

eeeXun:
We should get icons from `symbol_map`

### DIFF
--- a/lua/aerial/config.lua
+++ b/lua/aerial/config.lua
@@ -599,7 +599,7 @@ local function get_icon_provider()
   local has_lspkind, lspkind = pcall(require, "lspkind")
   if has_lspkind then
     return function(kind)
-      return lspkind.symbolic(kind, { mode = "symbolic " })
+      return lspkind.symbol_map[kind]
     end
   end
 


eeeXun:
1. in [this line](https://github.com/stevearc/aerial.nvim/pull/494/files#diff-f12c407f7430f790a3dc97ff3be8527ba2e5aeb62e1bd935ef9cf109f9f1610dL602). There is no mode **"symbolic "**. **"symbolic "** is not in [modes](https://github.com/onsails/lspkind.nvim/blob/3ddd1b4edefa425fda5a9f95a4f25578727c0bb3/lua/lspkind/init.lua#L97C1-L113C2). So in our current code, we will call [modes["symbol"]](https://github.com/onsails/lspkind.nvim/blob/3ddd1b4edefa425fda5a9f95a4f25578727c0bb3/lua/lspkind/init.lua#L173-L176).

2. In [modes["symbol"]](https://github.com/onsails/lspkind.nvim/blob/3ddd1b4edefa425fda5a9f95a4f25578727c0bb3/lua/lspkind/init.lua#L173-L176), it calls [get_symbol()](https://github.com/onsails/lspkind.nvim/blob/3ddd1b4edefa425fda5a9f95a4f25578727c0bb3/lua/lspkind/init.lua#L92-L95), and `get_symbol()` just returns `lspkind.symbol_map`.

> Is the symbolic() API not working for you?

Yes, it works. But it works by coincidence.

> Is there some reason why accessing symbol_map is better?

Yes, we don't need that recursive call.

BTW, the main reason of this PR is [nvim-cmp separates icons and kind](https://github.com/hrsh7th/nvim-cmp/pull/2190). So we no longer need `symbolic()` in lspkind. (I'm also planing create a PR here). And these changes will break this plugin. So, that's why I create this PR.

stevearc:
Not sure you'll be able to merge that change into lspkind, but the reasoning is sound. Happy to take the change here. Thanks for the PR!

PR #495, opened 2025-11-18
feat: support javascript prototype-attached fns

emmanueltouzery:


emmanueltouzery:
Hmm, I started to have doubts as to whether aerial should display the name as..

- alignCanvas
- PageElementWriter.alignCanvas
- PageElementWriter.prototype.alignCanvas

Maybe the last one makes more sense in the end? And it would also catch declarations like `x = function() {..}`. But I'll wait on your feedback to change it or not.

stevearc:
Hmmm...displaying it as simply `alignCanvas` is a bit misleading because then it looks the same as a `function alignCanvas()`. The last one `PageElementWriter.prototype.alignCanvas` is perhaps the most correct, but also takes a lot of horizontal space. I might lean towards `PageElementWriter.alignCanvas` as a balance of correctness and usefulness. What do you think?

emmanueltouzery:
I have no strong opinion either way. I think we can rule out the first/current one though.

Then I mentioned supporting the plain `x = function() {..}`, however I was confused, because you can't do that, then you need a `const` keyword, and then that's handled elsewhere already. 

So I'll do the version dropping the `prototype` as you slightly prefer it on practical grounds.

emmanueltouzery:
hmm but then is it possible at all to capture the class name and the method name and concatenate them? or to manually strip out the "prototype" in the treesitter scheme? maybe it has to be post-processed in lua then?

emmanueltouzery:
I decided to keep it simple and keep the prototype. Maybe it's slightly more correct, and a post-processing step seemed not warranted.

stevearc:
SGTM. Thanks for the PR!

PR #498, opened 2025-12-06
feat(fennel): add fennel support

ivanrg99:
Hi! Love the project!

I added fennel support for locals/vars and then function-like forms (lambdas, fns and macros), as well as tests.

Feedback is welcome

### DIFF
--- a/lua/aerial/backends/treesitter/extensions.lua
+++ b/lua/aerial/backends/treesitter/extensions.lua
@@ -404,6 +404,25 @@ M.zig = {
   end,
 }
 
+M.fennel = {
+  postprocess = function(bufnr, item, match)
+    local node = node_from_match(match, "symbol")
+    local full_form = get_node_text(node, bufnr)
+
+    if item.kind == "Function" then
+      local name = string.match(full_form, "[λ fn lambda macro] (%S+) %[")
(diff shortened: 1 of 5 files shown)


lua/aerial/backends/treesitter/extensions.lua
```diff
@@ -404,6 +404,25 @@ M.zig = {
...
+      item.name = name ~= nil and name or "<Anonymous>"
+    elseif item.kind == "Class" or item.kind == "Struct" then
+      local kind = (item.kind == "Class") and "local" or "var"
+      local query = kind .. " (%S+)"
+      local name = string.match(full_form, query) or "<parse error>"
+      item.name = name
+    else
+      item.name = "<parse error>"
+    end
+  end,
```
stevearc: Instead of using these extensions to extract the name, could you just adjust the treesitter queries? The `@name` capture group can be used to select the node that contains the name.

stevearc (changes requested):
Thanks for the PR! Left a couple of questions/suggestions

ivanrg99:
Closing this as I don't have bandwidth to work on this at the moment, and would rather not clutter the list of open PRs.

PR #509, opened 2026-02-23
feat(queries): update editorconfig

ValdezFOmar:
There were breaking changes to the grammar https://github.com/ValdezFOmar/tree-sitter-editorconfig/pull/56.

(Draft until [nvim-treesitter PR is merged](https://github.com/nvim-treesitter/nvim-treesitter/pull/8527))


stevearc (changes requested):
@ValdezFOmar I recently changed the symbol test snapshot format to be more readable. Could you rebase onto master and run `make update_snapshots`?

tests/symbols/editorconfig.snapshot
```diff
@@ -1,6 +1,7 @@
 Class *|4:1-4:2 (4:1-4:2)
-Class *.txt|11:2-11:7 (11:2-11:7)
-Class *.patch|11:8-11:15 (11:8-11:15)
+Class {*.txt,*.patch}|11:1-11:16 (11:1-11:16)
+  Class *.txt|11:2-11:7 (11:2-11:7)
+  Class *.patch|11:8-11:15 (11:8-11:15)
```
stevearc: This set of symbols seems a little noisy for representing a single header line
https://github.com/stevearc/aerial.nvim/blob/3b8d9a340b99f8e458ad8b689d69adc346472631/tests/treesitter/editorconfig#L11-L13

Is there any way we can modify the queries or postprocess the results so that we just have the top-level `{*.txt,*.patch}` symbol and not the two children `*.txt` and `*.patch`? I think that best represents the layout of the file, but open to other ideas.

stevearc (changes requested):

tests/symbols/editorconfig.snapshot
ValdezFOmar: > This set of symbols seems a little noisy for representing a single header line

I also though the same, but I was trying to preserve the logic of the previous queries. If you think is better this way, then just removing the second query is enough.

stevearc:
Awesome, thanks for the PR!

PR #510, opened 2026-02-23
FIX: Missing end_col value in asciidoc.lua causes errors on line 404 in window.lua.

coreyb-git:
Error introduced by correction in commit 68124b, causing "if" evaluation in windows.lua at line 404 to test against a nil value (the missing end_col value).

This bug was hiding if the cursor was at the leftmost position.

### DIFF
--- a/lua/aerial/backends/asciidoc.lua
+++ b/lua/aerial/backends/asciidoc.lua
@@ -37,7 +37,7 @@ M.fetch_symbols_sync = function(bufnr)
         lnum = lnum,
         col = 0,
         end_lnum = lnum,
-        end_col = 999,
+        end_col = line:len(),
       }
       if parent then
         if not parent.children then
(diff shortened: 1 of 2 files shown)


stevearc (changes requested):
@coreyb-git thanks for finding the issue, but instead of hard-coding 999 could we do something like `line:len()`?

Also, you may need to run `make update_snapshots` to regenerate the asciidoc symbol snapshot file for tests

coreyb-git:
Hi Steve, thanks for the tips.  I'm not familiar with the testing environment so you might need to do whatever needs doing to satisfy your test suite, though I have run the make command you asked.

You are more than welcome to close this pull request and just update that single line during your own daily tasks if that's easier for you.

stevearc:
Great, LGTM thanks!

PR #513, opened 2026-04-16
fix: use node:range() instead of deprecated node:start()

mirotorch:
In Nvim 0.12, node:start() has been deprecated, causing aerial to crash when calling it in helpers.lua. This PR replaces it with modern node:range() function.


qufeiyan:
It seems not to be working properly.

stevearc:
Thanks for the fix!

Issue #515, opened 2026-04-19
feature request: not tagging typescript type alias declarations as variables


hugolepeytre:
### Did you check existing requests?

- [x] I have searched the existing issues

### Describe the feature

Map typescript and tsx type alias declarations as Interfaces rather than Variables.

### Provide background

Right now Typescript types aliases are being tagged as kind "Variable", which makes sense because that's what they are.
But they behave almost exactly like Interfaces, so for the purposes of navigating code it seems like it might make more sense to tag them as such. I assume most users would want to see Interfaces and Types at the same time, without showing all of a file's variables.

### What is the significance of this feature?

nice to have

### Additional details

I made a PR (#516) in case the proposition gets accepted

PR #516, opened 2026-04-19
feat: treat ts and tsx type_aliases as Interfaces

hugolepeytre:
PR for #515 if proposition is accepted


stevearc:
Seems fine to me. Thanks!

PR #517, opened 2026-05-06
fix: update TSNode API for treesitter 0.12 compatibility

qufeiyan:
## Summary
- Fixes #513: `attempt to call method 'start'/'range' (a nil value)` in Neovim 0.12
- Root cause: `query:iter_matches()` in treesitter 0.12 returns captures as `table<integer, TSNode[]>` (array per capture), but the conversion loop stored the entire array as `.node` instead of extracting the first element
- Also replaces deprecated `:start()`/`:end_()` with `:range()` for future-proofing

## Test plan
- [x] Open aerial.nvim nav view on a buffer with treesitter support (lua, typescript, etc.)
- [x] Verify no errors about nil method calls
- [x] Verify symbols display correctly
- [x] Verify help file handling still works (extensions.lua change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lua/aerial/backends/treesitter/extensions.lua
+++ b/lua/aerial/backends/treesitter/extensions.lua
@@ -180,7 +180,7 @@ M.help = {
     local node = match.name.node
     if vim.startswith(node_from_match(match, "symbol"):type(), "h") then
       while node and node:type() == "word" do
-        local row, col = node:start()
+        local row, col = node:range()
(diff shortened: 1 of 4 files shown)


stevearc (commented):
The treesitter/init.lua and aerial.lua issues were fixed in f93dcee, so I'm going to merge #513 instead

PR #519, opened 2026-05-24
fix: don't crash when an aerial query file fails to parse

### LINKED ISSUE #506: bug: Latest aerial update error when opening .sql file
### Neovim version (nvim -v)

NVIM v0.11.5

### Operating system/version

MacOS 26.2

### Output of :AerialInfo

Aerial Info
-----------
Filetype:
Configured backends:
  treesitter (not supported) [No treesitter parser for ]
  lsp (not supported) [No LSP client found that supports symbols]
  markdown (not supported) [Filetype is not markdown]
  asciidoc (not supported) [Filetype is not asciidoc]
  man (not supported) [Filetype is not man]

### Describe the bug

After installing the latest aerial.nvim update, now when I open up a .sql file in nvim I get the following error.

Error executing vim.schedule lua callback: ...0.11.5_1/share/nvim/runtime/lua/vim/treesitter/query.lua:373: Query error at 37:2. Invalid node type "create_policy":
(create_policy
 ^

stack traceback:
        [C]: in function '_ts_parse_query'
        ...0.11.5_1/share/nvim/runtime/lua/vim/treesitter/query.lua:373: in function 'fn'
        ...im/0.11.5_1/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'fn'
        ...im/0.11.5_1/share/nvim/runtime/lua/vim/func/_memoize.lua:78: in function 'get'
        ...y/aerial.nvim/lua/aerial/backends/treesitter/helpers.lua:43: in function 'get_query'
        ...lazy/aerial.nvim/lua/aerial/backends/treesitter/init.lua:22: in function 'is_supported'
        ...share/nvim/lazy/aerial.nvim/lua/aerial/backends/init.lua:40: in function 'is_supported'
        ...share/nvim/lazy/aerial.nvim/lua/aerial/backends/init.lua:73: in function 'get_best_backend'
        ...share/nvim/lazy/aerial.nvim/lua/aerial/backends/init.lua:147: in function 'get'
        ...share/nvim/lazy/aerial.nvim/lua/aerial/backends/init.lua:251: in function 'attach'
        .../share/nvim/lazy/aerial.nvim/lua/aerial/autocommands.lua:88: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

Open up a file that ends in .sql

### Expected Behavior

Expect it to open the file without an error

### Minimal example file

The file can be blank or have a query in it

### Minimal init.lua

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/aerial.nvim",
    config = function()
      require("aerial").setup({
        -- add your aerial config here
      })
    end,
  },
  {
    build = ":TSUpdate",
    config = function()
        ensure_installed = { "c", "lua" },
        auto_install = true,
        highlight = { enable = true },
      })
    end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Additional context

_No response_

Booyaka101:
Closes #506.

`vim.treesitter.query.get(lang, \"aerial\")` raises whenever the query file references node types absent from the installed grammar. The current reported case: SQL files crash aerial with

```
Query error at 37:2. Invalid node type \"create_policy\":
(create_policy
 ^
```

because `queries/sql/aerial.scm` uses `create_policy`, which was added to tree-sitter-sql in v0.3.10 (2026-01) and is missing from older parsers (very common when using nvim-treesitter master branch on Neovim < 0.12 — see [DerekStride/tree-sitter-sql#347](https://github.com/DerekStride/tree-sitter-sql/issues/347)).

### The fix

Wrap `query.get` in `pcall` and return `nil` on parse failure inside `get_query`. Both existing callers already handle `nil` (`init.lua:22` in `is_supported`; `init.lua:184` in `get_lang_and_query`), so aerial just treats the language as treesitter-unsupported for that one user instead of crashing on every buffer enter. Emit a single user-facing warning; cache the failure so the warning doesn't spam on every buffer.

This is intentionally a defensive layer only — I'm not removing `create_policy` from `queries/sql/aerial.scm` since that was added on purpose in #505 and might be used by people on newer grammars. If you'd like to additionally restructure that file (e.g. split into a stable core + extras, or drop `create_policy` to keep partial outline working on old grammars), happy to follow up in a separate PR.

### How this was tested

End-to-end reproduction with a synthetic queries dir that references a non-existent node type for an installed grammar (lua):

| | Behavior |
|---|---|
| **Baseline** | \`FAIL: get_query raised: ...Query error at 5:2. Invalid node type \"__deliberately_missing_node_simulating_create_policy\"\` |
| **With fix** | \`aerial: failed to load treesitter query for lua, treating as unsupported...\` warning logged, \`get_query\` returns \`nil\` |

3 regression tests in \`tests/helpers_spec.lua\`:
- returns nil instead of crashing when the query fails to parse
- caches the parse failure so the warning only fires once (across repeated calls)
- still returns the query when parsing succeeds

All 3 pass with the fix; 2/3 fail on baseline (the success-case test passes either way).

### DIFF
--- a/lua/aerial/backends/treesitter/helpers.lua
+++ b/lua/aerial/backends/treesitter/helpers.lua
@@ -36,31 +36,21 @@ end
 
 ---@param lang string
 ---@return vim.treesitter.Query|nil
----@note caches queries to avoid filesystem hits on neovim 0.9+
+---@return string|nil err  parse error when the aerial query is invalid for this grammar
+---@note caches queries (and parse errors) to avoid filesystem hits on neovim 0.9+
 M.get_query = function(lang)
   if not query_cache[lang] then
-    -- pcall: a query file that references node types not present in the
-    -- installed grammar raises here. Without this, aerial crashes every
-    -- time the user opens a file in that language (see #506: SQL aerial.scm
-    -- references `create_policy` which is missing from older
-    -- tree-sitter-sql). Treat parse failure the same as "no query for this
-    -- language" so callers fall back gracefully.
-    local ok, query_or_err = pcall(vim.treesitter.query.get, lang, "aerial")
-    if not ok then
-      vim.notify(
-        ("aerial: failed to load treesitter query for %s, treating as unsupported. %s"):format(
-          lang,
-          tostring(query_or_err)
-        ),
-        vim.log.levels.WARN,
-        { title = "aerial.nvim" }
-      )
-      query_or_err = nil
+    -- Defensive against query files that reference node types missing from
+    -- the installed grammar. See #506.
+    local ok, query = pcall(vim.treesitter.query.get, lang, "aerial")
+    if ok then
+      query_cache[lang] = { query = query }
+    else
+      query_cache[lang] = { err = tostring(query) }
     end
-    query_cache[lang] = { query = query_or_err }
   end
-
-  return query_cache[lang].query
+  local entry = query_cache[lang]
+  return entry.query, entry.err
 end
 
 ---@param lang string
--- a/lua/aerial/backends/treesitter/helpers.lua
+++ b/lua/aerial/backends/treesitter/helpers.lua
@@ -39,7 +39,25 @@ end
 ---@note caches queries to avoid filesystem hits on neovim 0.9+
 M.get_query = function(lang)
   if not query_cache[lang] then
-    query_cache[lang] = { query = vim.treesitter.query.get(lang, "aerial") }
+    -- pcall: a query file that references node types not present in the
+    -- installed grammar raises here. Without this, aerial crashes every
+    -- time the user opens a file in that language (see #506: SQL aerial.scm
+    -- references `create_policy` which is missing from older
+    -- tree-sitter-sql). Treat parse failure the same as "no query for this
+    -- language" so callers fall back gracefully.
+    local ok, query_or_err = pcall(vim.treesitter.query.get, lang, "aerial")
+    if not ok then
+      vim.notify(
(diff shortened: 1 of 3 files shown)


stevearc (changes requested):
This has the look of minimally edited LLM output. Just from skimming the changes that I initially want to see are:
- remove the unnecessary, overly verbose comments
- instead of logging these errors, make them surface in the `:AerialInfo` diagnostics
- ensure that all tests are valuable and we're not just testing trivial logic

Booyaka101:
Thanks, pushed an update — errors now flow through `is_supported` so they show in `:AerialInfo`. Also trimmed the comments and dropped the trivial test.

stevearc:
LGTM, thanks for the PR!

PR #525, opened 2026-06-02
feat(cpp): add support for namespace symbols

Kacperos155:
Add support for `namespace` declarations in C++.
https://en.cppreference.com/cpp/language/namespace

Note: `Namespace` symbols are hidden by default.
image


stevearc:
Thanks for the PR!

